PR #250, opened 2024-10-27
Add CSRF environment variable for use behind reverse proxies

stignarnia:
I was using Authentik reverse proxy and I couldn't get it to work without changing django settings to allow my domain in CSRF


FuzzyGrim:
Hey, thanks for the pull request!

For a easier setup, we could append `localhost` to `ALLOWED_HOSTS` by default in the code and maybe use a common setting for `ALLOWED_HOSTS` and for `CSRF_TRUSTED_ORIGINS`, like in paperless:

```
# The next 3 settings can also be set using just PAPERLESS_URL
CSRF_TRUSTED_ORIGINS = __get_list("PAPERLESS_CSRF_TRUSTED_ORIGINS")

# We allow CORS from localhost:8000
CORS_ALLOWED_ORIGINS = __get_list(
    "PAPERLESS_CORS_ALLOWED_HOSTS",
    ["http://localhost:8000"],
)

if DEBUG:
    # Allow access from the angular development server during debugging
    CORS_ALLOWED_ORIGINS.append("http://localhost:4200")

ALLOWED_HOSTS = __get_list("PAPERLESS_ALLOWED_HOSTS", ["*"])
if ALLOWED_HOSTS != ["*"]:
    # always allow localhost. Necessary e.g. for healthcheck in docker.
    ALLOWED_HOSTS.append("localhost")


def _parse_paperless_url():
    global CSRF_TRUSTED_ORIGINS, CORS_ALLOWED_ORIGINS, ALLOWED_HOSTS
    url = os.getenv("PAPERLESS_URL")
    if url:
        CSRF_TRUSTED_ORIGINS.append(url)
        CORS_ALLOWED_ORIGINS.append(url)
        ALLOWED_HOSTS.append(urlparse(url).hostname)

    return url


PAPERLESS_URL = _parse_paperless_url()
```

stignarnia:
I did the first part, but for CSRF I don't know, because this doesn't want any IP or local things, only domain names with https on with the https:// part explicitly written

PR #377, opened 2025-03-16
fix: remove migrations that have been squashed in 03c813c

ajanvier:


FuzzyGrim:
Thank you for the PR! I plan to merge this after a few more releases to give users enough time to migrate their databases. If this is included in version 0.20.6, users upgrading directly from 0.20.4 to 0.20.6 would face issues. I'll hold off on merging for now to ensure a smoother upgrade path for everyone.

ajanvier:
Sure ! I was just having trouble reverting a migration I created because of the `0022_item_source` referring to the now-missing `0021_alter_item_media_type` but I guess it's just a development issue so not a big deal.

PR #392, opened 2025-03-22
feat: various search improvements

ajanvier:
- Use the current media type as search type per default (e.g. if we're in the Anime list, the search type is automatically set to Anime)
- Make the media list search placeholder more distinct from the header search
- Sort TMDB results (tv/movies) by popularity for more relevant results order

### DIFF
--- a/src/app/providers/tmdb.py
+++ b/src/app/providers/tmdb.py
@@ -28,7 +28,8 @@ def search(media_type, query):
 
         response = services.api_request("TMDB", "GET", url, params=params)
 
-        response = response["results"]
+        # Sort results by popularity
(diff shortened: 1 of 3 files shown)


FuzzyGrim:
Thanks for the PR, on sorting TMDB results by popularity, which queries did you find it useful? I've tried a few and for example with “Friends” or "Chernobyl", it seems that the default relevance sort seems to do better.

ajanvier:
> Thanks for the PR, on sorting TMDB results by popularity, which queries did you find it useful? I've tried a few and for example with “Friends” or "Chernobyl", it seems that the default relevance sort seems to do better.

I don't remember the exact shows I searched but I guess it's more of an API "issue" (to improve the results order) rather than something that can be handled in Yamtrack 😕

PR #415, opened 2025-04-05
feat: add release calendar export

### LINKED ISSUE #375: [Feature Request] Export the Release Calendar to iCalendar format
**Is your feature request related to a problem?**
No.

Export the release calendar as an iCalendar file (with static link) so it can be included in a Nextcloud or any calendar application.

Ideally it would be a link containing a unique token that could be easily accessed without need for additional authentication (which would not be supported by apps).

ajanvier:
Export the release calendar through an iCal file that can be downloaded from the calendar page or accessed through a link also displayed on the calendar page (authenticated using the user's token).

Fixes #375


FuzzyGrim:
Thanks!

FuzzyGrim:
Would there be any problem in changing from all day events to time specific events with time zone information? I can implement the change, but I would like to know if there would be any possible problems that I haven't thought of.

The current implementation creates day-long events using only the date component of release timestamps. This causes timezone discrepancies for users in different timezones, as UTC release times don't properly convert to the user's local date.

- Problem scenario example:
   - A TV episode releases at 2025-04-13 01:00 UTC
    - For a user in GMT-3 (Brazil), this is actually 2025-04-12 22:00 local time
    - Current implementation creates an all-day event on 2025-04-13
    - User misses the release because they're looking at April 13 when the show actually releases on their local April 12


```
# Current implementation (creates all-day events)
cal_event.add("dtstart", release.datetime.date())
cal_event.add("dtend", release.datetime.date() + timedelta(days=1))

# Proposed implementation (creates time-specific events with timezone)
local_dt = timezone.localtime(release.datetime)
cal_event.add("dtstart", local_dt)
cal_event.add("dtend", local_dt)
```

ajanvier:
If you want to use time-specific events, as the icalendar library extracts the timezone info from tz-aware datetime ([see docs](https://icalendar.readthedocs.io/en/latest/usage.html#property-parameters)), I would go with something like:

```python
dt_tz_aware = release.datetime.replace(tzinfo=timezone.utc)
cal_event.add("dtstart", dt_tz_aware)
cal_event.add("dtend", dt_tz_aware)
```

The calendar client will see that the time is defined as UTC and should adapt it depending on the localtime or user settings.
I haven't tried it though so please don't take my word for granted.

Issue #391, opened 2025-03-22
[Feature Request] Import from HowLongToBeat


Spencer-0003:
**Is your feature request related to a problem?**
No.

**Describe the solution you'd like**
The ability to import games from HowLongToBeat.

**Additional context**
HowLongToBeat does not offer an API but you can export your games via CSV.

Example export:  [HLTB_Games_2025-03-22.csv](https://github.com/user-attachments/files/19402487/HLTB_Games_2025-03-22.csv)

FuzzyGrim:
Added in [v0.23.0](https://github.com/FuzzyGrim/Yamtrack/releases/tag/v0.23.0)

PR #492, opened 2025-05-14
Add support for HowLongToBeat CSV import #391

Tusky:
Implemented functionality to import game data from HowLongToBeat CSV files. This includes a new form in the frontend, associated backend view, Celery task, and importer logic. Updated the template tags, URLs, and other relevant components to support this new integration.


FuzzyGrim:
Thank you!

Issue #408, opened 2025-04-01
[Feature Request] Plex integration


Nevarro:
**Describe the solution you'd like**  
Plex Webhook integration to track Movies and TV Shows. 

**Describe alternatives you've considered**  
-

**Additional context**  
Plex offers tracking via Webhooks (that's what I'm doing with Trakt atm. I'd love to use Yamtrack instead.

RealFascinated:
Would love to see this

PedroBuffon:
would be awesome, and if you don't pay plex pass, it could be a Tautulli integration

dsmitty166:
+1

throny:
Since Trakt changed their pricing (2-4x for VIP), this feature would make it easier to switch. I believe the userbase of Yamtrack will grow in the next weeks :)

binarymelon:
@FuzzyGrim If work hasn't begun on this, is it ok if I take a look?  This will be the 1 major feature I need to migrate away from Trakt.

FuzzyGrim:
@binarymelon I haven't started yet, it would be great if you could take a look at it. Here is the code for the Jellyfin webhook for reference:

https://github.com/FuzzyGrim/Yamtrack/blob/dev/src/integrations/webhooks/jellyfin.py

PedroBuffon:
@binarymelon I've already moved away from Simkl and its useless saas, i even deleted the account, now everything is on Yamtrack, only need this feature to make it perfect.

TheChrisK:
Just set up Yamtrack and cancelled my Trakt sub. Would love Plex integration as well. Thank you for this!

Danavans:
Same situation here, i am looking for a trakt replacement, but as a Plex pass user, i would love the webhook feature, with automatically watched history and rating directly from Plex app :) I have nothing against Jellyfin, but I prefer Plex.

cbackas:
If I don't see anybody working on this by the weekend I'll probably take a swing at it

ryck:
I am giving it a go, but Plex only exposes media.scrobble, so we will need to use that instead...
Also, plex doesn't give you the show guids, only the episode, so we need an extra request to find it.
But apart from that, I think is going well, I am already tracking tvshows successfully, next stop: movies, code cleanup, optimisation, etc...

Here is a draft PR: https://github.com/FuzzyGrim/Yamtrack/pull/516

psyciknz:
[...]

PR #535, opened 2025-05-25
Better error handling + ruff fixes

ryck:
Re: https://github.com/FuzzyGrim/Yamtrack/issues/408#issuecomment-2907140464

Tries to improve how we handle GUIDs in the Plex webhook, which is always a bit of a PITA because it's really hard to test all the edge cases, but now we should check for all the relevant GUIDs before giving up... Also, more guards, so error should be less likely.

In this PR I also fix all ruff (new discover for me, seems the prettier of python), so lot's of refactoring

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ You can try the app at [yamtrack.fuzzygrim.com](https://yamtrack.fuzzygrim.com) 
 - 🐳 Easy deployment with Docker via docker-compose with SQLite or PostgreSQL.
 - 👥 Multi-users functionality allowing individual accounts with personalized tracking.
 - 🔑 Flexible authentication options including OIDC and 100+ social providers (Google, GitHub, Discord, etc.) via django-allauth.
-- 🦀 Integration with [Jellyfin](https://jellyfin.org/), to automatically track new media watched.
+- 🦀 Integration with [Jellyfin](https://jellyfin.org/) and [Plex](https://plex.tv), to automatically track new media watched.
 - 📥 Import from [Trakt](https://trakt.tv/), [Simkl](https://simkl.com/), [MyAnimeList](https://myanimelist.net/), [AniList](https://anilist.co/) and [Kitsu](https://kitsu.app/) with support for periodic automatic imports.
 - 📊 Export all your tracked media to a CSV file and import it back.
 
--- a/src/app/providers/tmdb.py
+++ b/src/app/providers/tmdb.py
@@ -116,10 +116,9 @@ def find(external_id, external_source):
         except requests.exceptions.HTTPError as error:
             handle_error(error)
 
-
         cache.set(cache_key, data)
 
-    return response["tv_episode_results"][0] if response["tv_episode_results"] else None
+    return response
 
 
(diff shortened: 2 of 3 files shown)


nadnerb98:
Screenshot_20250528-035911
I'm using Yamtrack on the newest build that was just released on the Dev Branch. Trying to get the Plex Integration to work. I keep getting this error in my logs. I also have my Trakt Webhook in my account and that one works just fine. But nothing is popping up in my Yamtrack.
IMG_20250528_031131

ryck:
Logs are telling you what's happening:

The webhook is ignoring the event because you have the wrong config in the yamtrack plex integration config: nadnerb98

To fix it, find what is your username in plex (ie, go to your plex profile) and set that same user in yamtrack's plex integration settings...

nadnerb98:
Screenshot_20250528-054053
Screenshot_20250528-053924
That is my username. That's what I don't understand. Unless I need to leave the username blank since it's just my account. Maybe I'm just confusing myself lol.

FuzzyGrim:
Could you try again with the new build in the dev branch? I've added some logs, maybe with that we can figure out what's going on.

nadnerb98:
IMG_20250528_192048
Got the log. Looks like my username was different than my account said. Sorry guys. I'm smarter than I seem lol. It's working now. Thanks for y'all's responses

ryck:
The hook has the user `brendan`, but in yamtrack you have `nadnerb98` configured. They need to be the same in order for the hook to trigger...

nadnerb98:
> The hook has the user `brendan`, but in yamtrack you have `nadnerb98` configured. They need to be the same in order for the hook to trigger...

Yeah I saw after Fuzzy added those extra Username logs. For whatever reason even though Nadnerb98 is my account username, Berendan is my Profile name on Plex. So thats what was confusing Yamtrack lol. Thanks again guys for yalls help.

ryck:
To be honest, it's quite confusing, so we may need to add some more wording for it. In my other project I also accept user IDs (ie, a number) which is less ambiguous, but even harder to find...

nadnerb98:
> To be honest, it's quite confusing, so we may need to add some more wording for it. In my other project I also accept user IDs (ie, a number) which is less ambiguous, but even harder to find...

Yeah cause what it took wasn't my username. But granted most people probably have the same for the username and profile name too
Screenshot_20250529-082432

ryck:
> > To be honest, it's quite confusing, so we may need to add some more wording for it. In my other project I also accept user IDs (ie, a number) which is less ambiguous, but even harder to find...
> 
> 
> 
> Yeah cause what it took wasn't my username. But granted most people probably have the same for the username and profile name too
> 
> Screenshot_20250529-082432
> 
> 

Yeah, but sadly we don't get that "username" in the hook payload... that would be the best solution...

FuzzyGrim:
@ryck 

Is there any way to get the show ID from the `media.play` and `media.scrobble` payloads?

ryck:
No, no that I know of, that's a Jellyfin thing, and that's why I do an extra request to get it...

PR #548, opened 2025-05-28
Allow host under subpath using BASE_URL environmental variable

### LINKED ISSUE #472: [Feature Request] Serve under a subpath
I would like to serve yamtrack under a subpath, like: https://mydomain.com/yamtrack/

This practice is very popular in self hosted community, it reduces the number of subdomains to manage, reduces certificates and DNS to manage, also has some added privacy benefit because any observer on the line cannot see the subpath in https traffic. 

Using a subdomain which has the disadvantages listed above. 

**Additional context**  
I went through some documents and it seems to be relatively straight forward with django. Though I am not too familiar with python to contribute with a PR. 

Thank you!

christaikobo:
Fixes https://github.com/FuzzyGrim/Yamtrack/issues/472

Tested locally with default authentication flow and OIDC, seems to work without issue. 

User has to set a environmental variable `BASE_URL` for the docker container, and then use reverse proxy to strip the subpath. 

`BASE_URL` should be in the form of `/yamtrack`, without trailing slash. 

Not too familiar with python, so if there is any problem, please just modify directly. Thank you!

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -18,6 +18,10 @@ from decouple import (
 )
 from django.core.cache import CacheKeyWarning
 
+BASE_URL = config("BASE_URL", default=None)
+if BASE_URL:
+    FORCE_SCRIPT_NAME = BASE_URL
+
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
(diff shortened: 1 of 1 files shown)


FuzzyGrim:
Thanks, looks good to me!

PR #575, opened 2025-06-03
Adding GoogleBooks for book metadata

psyciknz:
Wasn't really sure what I was doing, so modeled it on openlibrary.  Some of the concepts a little different that openlibrary (in respect to a book or an edition).  They use volumes for a specific book.  But I wasn't sure of the link between searching for a title and if that gets back a volume (or something else where you then have to select a volume).

But hopefully I've done enough of the hard work, so that it's fairly easy to finish off.

Next time, I'll use a proper branch name as well.

### DIFF
--- a/src/static/css/main.css
+++ b/src/static/css/main.css
@@ -2020,6 +2020,16 @@
       }
     }
   }
+  .hover\:bg-gray-600\/50 {
+    &:hover {
+      @media (hover: hover) {
+        background-color: color-mix(in srgb, oklch(44.6% 0.03 256.802) 50%, transparent);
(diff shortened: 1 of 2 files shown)


FuzzyGrim:
Could you remove the extra files created? `pytest.init`, `launch.json`, `settings.json`, and the additional entry in `.gitignore`?

Thanks.

psyciknz:
that new pytest.ini was what was mentioned in here: https://github.com/FuzzyGrim/Yamtrack/discussions/513

But I've removed the launch.json and set setting.json back to official

PR #578, opened 2025-06-04
Implement Emby integration using webhooks

ryck:
At this point you may want to do some refactoring so we can share some functions between webhooks (maybe the ones related to yamtrack models, like to add an item to yamtrack)

Another option would be to use the same file, but get the parameter from the URL (ie, jellyfin|plex|emby) and work with that, but I don't particulary like that option, I like to have one file for each integration...

So maybe having some common functions is the best approach?

Also, do you think there is interest for a Tautulli integration, so you don't need a Plex Pass to use webhooks?


FuzzyGrim:
Yes, I'll refactor it after the merge. Could you add some tests with sample payloads so I can work with it?

Thanks.

FuzzyGrim:
I have done some refactoring in `dev`, all the tests are passing but if possible, could you verify that both Plex and Emby webhooks are working as expected?

> Also, do you think there is interest for a Tautulli integration, so you don't need a Plex Pass to use webhooks?

Yes, can it copy exactly the Plex Webhook payload format? If so, we could use the same end point, right?

I'm thinking of releasing `v0.24.0` soon, do you think we could add the Plex webhook handler for the `rate` event?

Issue #397, opened 2025-03-25
[Feature Request] Music Tracking


ksharizard:
**Describe the solution you'd like**  
I would like a way to track music. It should be able to rate albums, singles, artists and more. There are APIs to get the data from like discogs, spotify or lastfm. 

**Describe alternatives you've considered**  
I have used spotify playlists to track music but that was inefficient. I also used ryot's music tracking but it is in its preliminary stages.

dannyvfilms:
I decided to work on a very basic music‐tracking implementation with the existing Plex webhook as the source. 

✅ Completed so far
- Created new Artist, Album, and Track media types in the database and models.
- Added visibility toggles in Sidebar Settings to appears alongside TV, Movies, etc.
- Clicking “Artists”, “Albums”, or “Tracks” in the sidebar now correctly lists any logged items.
- Adjusted Plex webhook logic so that when new music plays, it automatically creates or updates the corresponding Track record.

![Image](https://github.com/user-attachments/assets/a39a122b-5515-41eb-8618-8dcaeea32b36)
![Image](https://github.com/user-attachments/assets/5595d17c-ec87-440b-8a84-18ac8d4b0c30)

⏳ Stuck on
- Nesting logic: we need “Tracks → Albums → Artists” exactly the same way TV uses “Episodes → Seasons → Shows.”
- Right now “Tracks” are flat; they don’t yet attach under their parent Album.
- Likewise “Albums” exist independently but aren’t grouped under their parent Artist.
- Unless o3 can figure this out in a way that o4-mini can't, I will leave this as-is for someone else to fix. I'm reasonably confident that the associations are logged on track creation, it's just displaying them I'm struggling with.

🚧 Planned next steps
- Explore integrating a metadata API (e.g. Last.fm) to pull album art, release dates, track lists, etc., to clean up the visuals.
- I haven't tried this step yet, so there is a chance it is beyond my ability to implement.

❌ Not in scope for this initial MVP
- Support for other scrobble input sources (e.g. Spotify, Last.fm scrobbling) beyond Plex.
- Any advanced tasks to import library contents similar to Trakt importing.

I don't have a background in coding. I just realized that if I could duplicate enough of the existing code and match it as closely as possible, much of the groundwork could be laid quickly (about 2 days). The polish will come from skills I don't have, so I'm seeing how far o3 can take me to add a feature I'd love to have as well. I’ll need help ironing out the details and polishing the UI/UX to match the quality of the other features.

[...]

PR #604, opened 2025-06-12
Collapse Recommendations by Default

### LINKED ISSUE #599: [Feature Request] Collapse recommendation by default
Currently there is a recommendation section under each entry. Maybe occasionally I would want to look at some recommendations, but most of the time they just took a lot of space and are distracting. 

Currently there does not seem to be options related to the details page, maybe collapse it by default would be preferreble. 

Thank you!

I try to ignore it but it's too distracting.

christaikobo:
Closes https://github.com/FuzzyGrim/Yamtrack/issues/599

With this, the recommendations at the bottom is collapsed by default.

### DIFF
--- a/src/templates/app/media_details.html
+++ b/src/templates/app/media_details.html
@@ -637,12 +637,27 @@
         {% for name, related_items in media.related.items %}
           {% if related_items %}
             <section class="{% if not forloop.last %}mb-8{% endif %}">
-              <h2 class="text-xl font-bold mb-4">{{ name|no_underscore|title }}</h2>
-              <div class="grid grid-cols-[repeat(auto-fill,minmax(150px,1fr))] gap-4">
-                {% for item in related_items %}
-                  {% include "app/components/media_card.html" with item=item title=item.season_title|default:item.title media=None %}
-                {% endfor %}
-              </div>
+              {# Conditionally apply <details> for 'Recommendations' #}
+              {% if name == "recommendations" %}
+                <details> {# NO 'open' attribute, so it's collapsed by default #}
+                  <summary class="text-xl font-bold mb-4 cursor-pointer py-2 px-2 rounded hover:bg-gray-100 active:bg-gray-200 transition">
+                      {{ name|no_underscore|title }}
+                  </summary>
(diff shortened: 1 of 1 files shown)


FuzzyGrim:
Hi, thanks for your contribution!

I think it would be better to have a toggle in settings to hide it by default because other users may not like hiding it by default. Do you think you can do that?

christaikobo:
Hi thanks for the suggestion. I can try.

There does not seem to be an appropriate place for such a setting right now?

dannyvfilms:
Probably best to add a new category called Preferences to Settings in the same area as Account, Notifications, Sidebar, etc. I've had some thoughts on optional UI features I've considered drafting and would also need an area to be able to toggle them on and off.

christaikobo:
@dannyvfilms Do you intend to open a PR in the near future for the preference tab in setting menu? If so, I think you can simply include this in your PR, and I will close this one. My PR is quite simple with minimal code changes in one place only.

If it is going to take a couple of weeks then I might try my luck with it.

dannyvfilms:
I haven't started anything in this area yet. Right now I'm trying to figure out if my help on Music Tracking #397 should be a PR right now, or if I should get help to finish it first before doing so. I am not a developer, have only AI assisted experience, and have only changed or added what I could use existing code reference to figure out. So since my existing changes don't include additional settings in the scope, it probably won't happen in what I have cooking right now.

christaikobo:
I have tried semi hard, and it seems to be a bit more complicated than I am able to handle. @FuzzyGrim it's up to you how to proceed. Should I close this one for now?

FuzzyGrim:
I'll need a preferences tab for #606, so you can try again after I implement that, but if you feel it's too much, I can handle this too. I'm fine with either way :)

christaikobo:
It would be great if you can implement https://github.com/FuzzyGrim/Yamtrack/issues/606 and this at the same time.

If you decide to implement https://github.com/FuzzyGrim/Yamtrack/issues/606 first, I'll have a look again later, maybe I can figure it out, maybe not... Either way we'll have to wait for that first.

PR #623, opened 2025-06-18
Configure SESSION and CSRF cookie for subpath

christaikobo:
Follow up on https://github.com/FuzzyGrim/Yamtrack/pull/548

I think it would be a good idea to configure cookies in subpath if subpath is configured.


FuzzyGrim:
Thank you, but next time please use the ruff formatter :)

Issue #629, opened 2025-06-18
[BUG] CSV import stops on a failure


daften:
**Describe the bug**
I have converted Trakt data (export) to the Yamtrack format with some scripts. I notice with this that the import fails if an item is e.g. not found on tmdb. One item had a tmdb movie id "1018501" but that id no longer exists on tmdb. When the import reaches that item ,the import stops.
This could happen with a yamtrack export too of course.

I entered this as a bug, this could also be considered a feature request

**To Reproduce**
Use a CSV with content that includes this header and line:
```
media_id,source,media_type,title,image,season_number,episode_number,score,status,notes,start_date,end_date,progress
1018501,tmdb,movie,,,,,,Completed,,,"2021-12-29 13:11:02+00:00",
```

**Expected behavior**
The import continues or there is an option to stop on failure or continue.

**Yamtrack version**: 0.24.2
**Database**: SQLite

FuzzyGrim:
Thanks, I agree. The import should continue with a warning message.

Is there a reason you're converting Trakt data to Yamtrack format if you can import directly from Trakt?

daften:
I did it because the trakt import didn't complete 100%, it's a lot of data, so I think it also stopped at some point. That or it's just that some seasons aren't automatically marked as completed even if all season episodes are completed.

A second reason is also that I'm experimenting with AI-driven development and creating scripts that do more advanced transformations is part of that. The bugs I notice in between I now also try to fix with AI. It helps me increase my skills with AI and hopefully helps projects like this as well :)

FuzzyGrim:
Released in [v0.24.3](https://github.com/FuzzyGrim/Yamtrack/releases/tag/v0.24.3)

daften:
Thanks again!
I noticed one more issue, but I'll try to create a fix myself when I have some time.
When it's a show that can be found, but an episode that doesn't exist on tmdb, e.g. cheers season 10 episode 26 (it does e.g. exist on tvdb and trakt), there is no services.ProviderAPIError thrown, but the error is 
```
row["title"] = metadata["title"]
                   ~~~~~~~~^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

I'll take a look what the best solution is :)

PR #630, opened 2025-06-18
Issue #629: Don't stop yamtrack import on error

daften:
A fix for issue #629 
For full transparency: This was generated with the help of Claude Sonnet, I'm not a python developer.

### DIFF
--- a/src/integrations/imports/yamtrack.py
+++ b/src/integrations/imports/yamtrack.py
@@ -66,8 +66,10 @@ class YamtrackImporter:
             try:
                 self._process_row(row)
             except Exception as error:
-                error_msg = f"Error processing entry: {row}"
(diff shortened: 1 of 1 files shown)


FuzzyGrim:
Thank you!

I will keep:

```
except Exception as error:
    error_msg = f"Error processing entry: {row}"
    raise MediaImportUnexpectedError(error_msg) from error
```

Because if the user has used an unexpected CSV format, it should stop the import so we don't make unnecessary API requests for each media ID when they won't be able to be imported anyway.

daften:
I’ll tackle that in a separate PR tomorrow normally!

FuzzyGrim:
It's fine, I will do it. Thanks

PR #642, opened 2025-06-19
Fix episode not found on import stopping import

daften:
This fixes the additional item I mentioned in https://github.com/FuzzyGrim/Yamtrack/issues/629#issuecomment-2988970568

I updated and added automatic tests that should capture the changed behaviour too. I placed the error in the episode method, that seemed the best place to me, but  I can easily move it to e.g. the get_media_metadata method.

### DIFF
--- a/src/app/providers/manual.py
+++ b/src/app/providers/manual.py
@@ -1,5 +1,6 @@
 from app import models
 from app.models import MediaTypes, Sources
+from app.providers import services
 
 
 def metadata(media_id, media_type):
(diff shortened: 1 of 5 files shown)


FuzzyGrim:
Thank you for addressing that issue!

The `providers/manual.py` is for manually created entries (https://yamtrack.fuzzygrim.com/create), the file you should change is `providers/tmdb.py`. Also instead of creating `services.MediaNotFoundError` I would prefer to trigger `services.ProviderAPIError` as it's already handling other cases of media not found.

daften:
I've rebased my changes on top of dev and I've made the changes as requested, not using a new Error, but the existing Error and putting it in tmdb.py! Let me know how this looks to you! :)

FuzzyGrim:
Thanks!

PR #708, opened 2025-07-08
Make logout redirect URL configurable through environment variable

p7996619:
Currently, when using OpenID-Connect for authentication, logging out does not redirect to the authentication provider. Hence, the user is not actually logged out. On page reload the user is directly logged in again as the cookie still exists.
In some cases this might be desired behaviour, but I assume more often than not it is expected to log out the user at the external auth provider. This can be easily achieved though by manually setting the `ACCOUNT_LOGOUT_REDIRECT_URL` value to the absolute URL of the auth provider logout endpoint. Example for Keycloak:
```
https://oauth.internal/realms/master/protocol/openid-connect/logout?post_logout_redirect_uri=/accounts/login/?loggedout=1&client_id=yam.internal
```
Ideally, django-allauth would parse and use this endpoint automatically from the OpenID discovery endpoint, but this doesn't seem to be implemented right now. see https://codeberg.org/allauth/django-allauth/issues/3104

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -4,7 +4,7 @@ import json
 import warnings
 import zoneinfo
 from pathlib import Path
-from urllib.parse import urlparse
+from urllib.parse import urljoin, urlparse
 
 from celery.schedules import crontab
 from decouple import (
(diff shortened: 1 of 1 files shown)


FuzzyGrim:
Thank you!

Issue #451, opened 2025-04-30
[Feature Request] Steam Integration 


JoghurtSommelier:
**Describe the solution you'd like**  
A Steam integration to automatically pull playtime, achievements etc.

**Additional context**  
Documenting the playtime of games (I may be still playing) is a hassle. Steam serves an API which can pull a lot of information by just using the users SteamID and can be used to pull information like playtime to automatically integrate it.

FuzzyGrim:
Added in the latest release.

PR #711, opened 2025-07-08
Inital Implmementation of Steam Game import functionality

ffejhog:
Implemented an initial version of Steam game importing. I followed the pattern used for other imports (ex Kitsu).

A new env variable was added STEAM_API_KEY for a steam api key (can be retrieved at https://steamcommunity.com/dev/apikey). I didn't include a default one, though it should be easy enough to add if desired. 

This functionality uses the GetOwnedGames API to get the list of games from the user, then matches them to IGDB (with a fallback to manual creation). This requires your steam profile to be public, you also must provide a SteamId64 (which can be gotten though your steam profile or sites like https://www.steamidfinder.com/

If you have a lot of Steam games there can be rate limiting with IGBD so I add retries with backoff to try to reduce timeout issues.

As a future update, I believe there would be value in a new "Update" import mode that simply updates the play time, ratings, etc for items.

Addresses #451


FuzzyGrim:
Hi, thanks! Looks great, just one thing: I think instead of searching IGDB by title, we can use the [External Game](https://api-docs.igdb.com/#external-game) endpoint, which allows you to find an IGDB ID by a Steam ID:

```
-d fields game; where uid = "271590" & external_game_source = 1;'
```

ffejhog:
Ah Good point. Just implemented external_game in the IGDB provider. I also added an enum for all valid external game sources if others are implemented in the future.

FuzzyGrim:
Thank you! Could you fix the ruff errors?

ffejhog:
@FuzzyGrim Done, ruff errors should be fixed and the merge conflict with the new imdb import feature was resolved.

FuzzyGrim:
Thank you!

Issue #534, opened 2025-05-24
[Feature Request] Import from private trakt profile


bermeitinger-b:
I would like to import my Trakt data without the need to set my profile to public.

This could be done with an API key. I already do have such a key, so I wouldn't mind adding it in a "advanced user settings" option.

psyciknz:
What about import via the backed up json files?  If you run their backup (to drop box)m, or request them for all your data, it';s the same.  Just needs no api acesss as they use an oath - which may or may not have been implemented

There's a movies.json and shows.json, but they only hold the latest watch of each.
Then a bunch of historyx.json files which have the complete watch history.

bermeitinger-b:
I'm fine with any import. csv, json, whatever, but it should not require too many manual steps. The (authenticated) API brings all data for importing.

psyciknz:
yeah I've just been playing with the trakt api.  I found the trakt.py project (version 4.4 which almost every trakt based project uses) wouldn't actually authenticate any more.  But I've manage to cobble a fix for it.

Currently reading all items from my trakt account.  Then its seeing if I can work out the yamtrack forming form importing (I'm currently dumping into influx which I run here).  I'm not saying I'll look at this but at least might get some parts that might help with it.

rantanlan:
And here I am struggling to get this working for hours with a private trakt profile… Thanks for your report!
maybe you can add a note to the wiki @FuzzyGrim ? For my understanding it is missleading that I need a public profile, while providing a api key or do I miss something ?

FuzzyGrim:
The API key is required to use the Trakt API. I provide a default API in the app, so it's optional, but providing your own is recommended for rate limiting reasons.

To access a private profile using the API, Trakt requires a secret key and requires an OAuth process, which I haven't implemented yet. The same applies to the other import providers: MyAnimeList, AniList, Kitsu.

But I do think it's a good idea to add a note, I'll add it to the app itself, thanks!

FuzzyGrim:
Added in the latest release.

PR #753, opened 2025-07-27
feat: add oauth import for trakt

baswag:
Added OAuth possibility for importing data from Trakt, with new env `TRAKT_API_SECRET` needed as configuration
This will make is possible to import data from private profiles (#534)

I left the non-OAuth code for now for compatibility reasons, users migrating to the new Oauth Importer will have to delete their Trakt import Jobs and recreate them using OAuth.


CrazyTim71:
Very nice! 

Perhaps this could be used as a basis for Anilist too? It would be great if data from private profiles could be imported from there too. I'll definitely take a look at it as soon as your PR has been merged :)

PR #770, opened 2025-08-02
fix: trakt refresh_token expiring

baswag:
fixed bug where trakt refresh_token was not updated after usage leading to no longer being able to get a new access_token refactored tasks and importers to use **_ for unused arguments


FuzzyGrim:
Thanks for the PR!

I will close it because I already commited a fix at https://github.com/FuzzyGrim/Yamtrack/commit/33b29440d67f204d0214634bc85af3a285052758

Issue #488, opened 2025-05-14
[Feature Request] REST API


fhemberger:
**Describe the solution you'd like**  
It would be nice to have a REST API to write own integrations or use the library data in other contexts (e.g. build a "recently watched/read" widget for your own website).

Useful endpoints to start:
- List items in category (e.g. `/medialist/tv`)
- Statistics: Last X watched/read items. Basically the existing media timeline, but also filterable per category
- Create custom entry

sahara101:
Yes, please. API and Plex webhook is what I need to cancel trakt. I checked the demo and it looks like the lists already do what I need.

hcgiub001:
Now that TRAKT have suddenly changed pricing structure, a real possible threat looms over that scrobbling will be taken off free tier so a service like this is essential to fallback if/when that time comes.
An API would potentially make this viable for kodi community which are in the millions.
This is something amazing you created and hope everyone involved is able to make this happen THX

umbrellaplug:
Hello, I'm a developer for an addon for kodi. I have been looking at different solutions for tracking watch history and stop and resume points across devices. I wanted to comment on this request to see if it would be possible to get api endpoints to allow for adding a show/ movie to watch history, removing a show/ movie from watch history, get all watch history, add stop point for movie or episode, remove stop point for movie or episode, get all saved stop points, and an activity endpoint that would give the time that each one of these were last updated. If you need clarification or if this is something that's possible please let me know.

somedevreally:
@FuzzyGrim if these are added, trakt will be history :) thankyou for your interest @umbrellaplug 

- watchlist
- watch history 
- stop and resume points

Enissay:
An API is the key to everything indeed: Extensions, Mobile APP, ...

pplcunha:
> @FuzzyGrim if these are added, trakt will be history :) thankyou for your interest @umbrellaplug 
> 
> - watchlist
> - watch history 
> - stop and resume points

And rating?

dmryce:
This is the feature I'm most waiting for. Due to the price increases at Trakt, I'm looking to switch. Yamtrack is the best alternative I've found.

javikr:
Hi there! I'm the developer of Sofa Time (an iOS app for tracking movies and TV shows). The app can sync with Trakt, Simkl, and iCloud (local). Adding a public REST API for Yamtrack would be a great feature, and I’m quite sure we would integrate it into the app if possible.

[...]

PR #772, opened 2025-08-03
implement a rest-endpoint for medialist

chinatsu:
i was a little inspired by #488 and decided to spend a little while to make a very rudimentary endpoint for medialist.

the api key used for other integrations is what authorizes the call, though i stuck it into a query parameter rather than part of the url like the other integrations, because i didn't really want something looking like `/api/medialist/tv/apikeyhere`

i ran into some issues trying to set up every data provider (especially mal), so i haven't been able to test it properly-properly.

during my own testing, i tried using these endpoints in my *arr-stuff, just to see if i could use the endpoint as an import list, and import a tv series and a movie. this is my main motivation to be working on this (hence the camelCase keys.. lol). i believe these integrations ignore unfamiliar keys, so we could definitely expand on the return types, and make it more unified. e.g. this for every media type?

```python
{
    "id": media_id,
    "tvdbId": None if not tvdb_id else tvdb_id,
    "name": name,
    # etc.
}
```

i'll look into this a little bit, though i'm happy to receive feedback, and i'm also fine having this closed if it's super stupid to add to your codebase!

### DIFF
--- a/src/integrations/tests/test_api.py
+++ b/src/integrations/tests/test_api.py
@@ -1,0 +1,179 @@
+import json
+from datetime import UTC, datetime
+from urllib.parse import urlencode
+
+from django.contrib.auth import get_user_model
+from django.db.models import Q
+from django.test import TestCase
+from django.urls import reverse
+
+from app import providers
+from app.models import (
+    Book,
+    Episode,
+    Item,
+    MediaTypes,
+    Movie,
+    Season,
+    Sources,
+    Status,
+)
+
+class APITest(TestCase):
+    """Test getting items from the api."""
+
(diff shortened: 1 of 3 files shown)


FuzzyGrim:
Isn't the response very slow because you are fetching metadata for each media?

chinatsu:
i haven't tested this branch on any large set of data, so it hasn't been particularly slow for me. i'll give that a shot though, and see how slow it is!
the use case for me requires tvdbId though, but maybe i could conditionally fetch that when it's needed. :thinking:

FuzzyGrim:
I will close this PR in favor of #924

PR #773, opened 2025-08-03
AniList: Add Oauth import for private profiles

CrazyTim71:
This PR add Auth support for the AniList import. Now one is able to import data from a private profile :D

To make it work, you need to create an app in your AniList profile settings. You'll get an ID and Secret to place into the .env file:
```
ANILIST_ID=...
ANILIST_SECRET=...
```

Go to [https://anilist.co/settings/developer](https://anilist.co/settings/developer)

grafik

Unbenannt

Screenshot 2025-08-26 202608

Screenshot 2025-08-26 204043

grafik

grafik

**Questions before merging:**

1. Do we need a migration for existing periodic tasks? If yes, how do we get the token?
2. How do we handle expired tokens since AniList doesn't support refresh tokens yet. Theoretically it would be enough to recreate the periodic import task by reauthenticating.
3. What about the import test for AniList? Public imports won't work anymore.
4. Should we add a wiki entry on how to configure the AniList App? It's neccesary to make this work.

Big Thanks to @baswag for creating the original PR which my code is heavily based on!


FuzzyGrim:
Hey, thanks for the PR!

> How do we handle expired tokens since AniList doesn't support refresh tokens yet. Theoretically it would be enough to recreate the periodic import task by reauthenticating.

Yes, I think the only way is to reauthenticate. The token expires after a year, so it's not too inconvenient.

> Should we add a wiki entry on how to configure the AniList App? It's neccesary to make this work.

Yes, I will add a wiki entry page on how the different imports work.

> Do we need a migration for existing periodic tasks? If yes, how do we get the token? What about the import test for AniList? Public imports won't work anymore.

We can't generate the tokens for existing tasks so I think we should allow both public import and Oauth methods for now.

CrazyTim71:
@FuzzyGrim Thanks for your feedback!

> Yes, I think the only way is to reauthenticate. The token expires after a year, so it's not too inconvenient.
> 
I actually didn't know that the token will last THAT long lol. But thats nice for us 😄
> Yes, I will add a wiki entry page on how the different imports work.
> 
Perfect, thanks!
> We can't generate the tokens for existing tasks so I think we should allow both public import and Oauth methods for now.
> 
Okay, I'll update the PR to allow both public and private imports. I did actually think of this at the beginning, but I saw that @baswag had replaced the public import for Trakt and Simkl with the private one. Wouldn't it therefore make sense to support both import methods for them too?

FuzzyGrim:
> Wouldn't it therefore make sense to support both import methods for them too?

Yes, I'll re-add the public import for Trakt. Simkl never had it because they don't support it.

CrazyTim71:
@FuzzyGrim Perfect, thanks!

I also added the public import for AniList to this PR, so both work now. 
I would love to hear your feedback on this :D

PR #776, opened 2025-08-04
fix tab focus for the login and the signup page

CrazyTim71:
This PR makes the login and signup process smoother. Its kinda annoying to tab over the eye icons 😅

**Before**:
before

**After**:
after


FuzzyGrim:
Thank you 👍

PR #813, opened 2025-08-15
Display runtime for each episode

### LINKED ISSUE #804: [Feature Request] Show runtime for each episode
Currently episode runtime is only displayed for the season. I assume this is the normal or average runtime. But runtimes can vary a lot between episodes, so it would be useful to see the runtime for each episode individually.

andrebk:
Display the runtime on each episode in a season overview. Since runtimes may differ a lot over a season, this gives a better overview than just the average runtime of the season.

Styling looks a little weird on mobile, since the line is now longer and sometimes needs to fold. But there are already issues with styling on small screens (can't read full episode titles in many cases), so leaving that for another PR.

<details>
<summary>Mobile styling examples</summary>

360px wide (Galaxy S20):
image

414px wide (iPhone 11 Pro Max):
image

</details>

Closes #804.

## Screenshots
Before:
image

After:
image

### DIFF
--- a/src/app/providers/tmdb.py
+++ b/src/app/providers/tmdb.py
@@ -549,7 +549,7 @@ def process_episodes(season_metadata, episodes_in_db):
                 "title": episode["name"],
                 "overview": episode["overview"],
                 "history": tracked_episodes.get(episode_number, []),
-                "runtime": episode["runtime"],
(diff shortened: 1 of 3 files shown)


FuzzyGrim:
Thanks!

PR #834, opened 2025-08-24
Support setting Redis prefix using environmental variable REDIS_PREFIX

christaikobo:
Redis does ACL using key and channel prefixes, if multiple apps and services were to use the same Redis instance as cache, the only way to enforce some kind of isolation and permission control between these apps and services is to set appropriate prefixes in them and control the permission in Redis. 

This PR introduces a environmental variable REDIS_PREFIX, and adds corresponding prefixes for Django, Celery and Yamtrack's own session keys and channels.

### DIFF
--- a/src/app/providers/services.py
+++ b/src/app/providers/services.py
@@ -34,10 +34,13 @@ def get_redis_connection():
 
 redis_pool = get_redis_connection()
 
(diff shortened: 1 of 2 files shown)


FuzzyGrim:
Thanks for the PR!

PR #842, opened 2025-08-27
Show tracking status on more media cards

### LINKED ISSUE #814: [Feature Request] Show tracking state icon on more cards
On the main page for each media type (Movies, TV Shows, etc.) the card for each item has an icon in the top left to show its tracking state (planning, completed, etc.). There is also a date, if eligible. But there are many other places where cards are shown for a media item where this icon is missing:

- Seasons on the page for a TV show
- Recommendations
- Search results

This makes it hard to quickly filter out which information is relevant without clicking into the item and checking its state.

I suggest showing the state icon in these locations as well. Often these cards are smaller than on the main pages, so showing the date as well is maybe overkill. Along the lines of what is shown when an item is planned:

Image

Current way of checking state is to click into the item, which is cumbersome. An alternate feature could be to show state on hover, but that would not be very scannable, and would also clash with expected behaviour from main pages where hover is used to show edit actions.

tuur29:
closes https://github.com/FuzzyGrim/Yamtrack/issues/814

## Screenshots

<details><summary>In search results</summary>
<p>
image
</p>
</details> 

<details><summary>In seasons overview & recommendations</summary>
<p>
image
</p>
</details> 

<details><summary>Also for other types of media (example games)</summary>
<p>
image
</p>
</details>

### DIFF
--- a/src/app/helpers.py
+++ b/src/app/helpers.py
@@ -6,6 +6,7 @@ from django.shortcuts import redirect
 from django.utils.encoding import iri_to_uri
 from django.utils.http import url_has_allowed_host_and_scheme
 
+from app.models import BasicMedia, Item, MediaTypes
 
 def minutes_to_hhmm(total_minutes):
     """Convert total minutes to HH:MM format."""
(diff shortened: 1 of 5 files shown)


FuzzyGrim:
Thanks for PR!

PR #869, opened 2025-09-14
Refactor SVG icons to separate files

andrebk:
## Summary
Most icons are currently included as `<svg>` elements directly in the templates. This has a number of drawbacks:

- Poor readability, since you can't tell at a glance which icon the svg code represents
- Duplicate code, since icons are redefined for each time they are used
- Hard to maintain, changing an icon means making changes in multiple places

This PR moves most icons to separate svg template files, which can then be included in other templates. 

The svg templates can be customized with classes by providing to them to the template when including it. This has some drawbacks:

- Cannot use template variables (`{{ }}`) or conditionals (`{% %}`) in the classes, since those can't be used inside the `{% include %}` statement. For the most part this can be solved by wrapping the icon in a `div` and styling that.
- The same goes for htmx attributes like `x-show`, will now have to be set on a wrapping element.

## Notes
I noticed the colors for the different tracking states come from the media type colors (e.g. `MediaTypes.MOVIE.value|media_color` for the paused state). Shouldn't this be two separate things?

There is one icon I wasn't able to refactor (the stars when setting rating), because it used `:class` to dynamically decide if the icon should be filled or not. Since the fill CSS property isn't inherited by children, I couldn't set it on a wrapping element instead, like the other places with `:class`. If anyone has a good solution for this please let me know.

Some icons were visibly identical, but had slightly different svg code. These have been unified.


FuzzyGrim:
Thank you for the PR!

For the `:class` issue, I decided to change the SVG `fill` property to `fill="currentColor"`.

Issue #801, opened 2025-08-12
[Feature Request] Anime Alternative Title


bhdouglass:
**Describe the solution you'd like**  

It would be nice to have an option to display the "alternative title" for an anime. It looks like the MAL api has alternative titles available in English and Japanese. 

**Describe alternatives you've considered**  

N/A

**Additional context**  

N/A

UptightGlitter:
That would be great! Really hope it gets implemented

vfnm:
Would also appreciate it.

Elderflowa:
I agree.

Navigating the anime category in japanese is a challenge. English titles would be a big help

NoPlayerD:
i agree, hope it'll be implemented soon

PR #890, opened 2025-09-28
Optional settings for MAL titles

marc-roston:
Addresses #801 

Introduces a new environment variable MAL_PREFER_EN_TITLE with default set to False, making this an opt in feature so as to not affect existing setups.

Titles in Media List and Media Details will automatically use the English titles from the MAL API, if they are available.

This PR does not affect Recommended or Related lists as the API response does not contain alternate titles. Manually querying each title in the Recommended and Related lists results in rate limiting from the MAL API which leads to a slow user experience and probably frowned upon by MAL, so decided against including it, even as an option.

image
image

image
image

### DIFF
--- a/src/app/providers/mal.py
+++ b/src/app/providers/mal.py
@@ -72,7 +72,7 @@ def search(media_type, query, page):
                 "media_id": media["node"]["id"],
                 "source": Sources.MAL.value,
                 "media_type": media_type,
-                "title": get_english_title_if_available(media["node"]),
+                "title": get_title(media["node"]),
                 "image": get_image_url(media["node"]),
             }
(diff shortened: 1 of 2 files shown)


dakoellis:
Hey @marc-roston I pulled your PR into current dev and it seems to work as expected for new titles, however pre-existing titles are in a weird state.  They have the english titles when viewing the full details, but in the lists on home/anime tabs, they have the Romanized Japanese names. Any chance you know of a way to fix it without having to manually remove and re-add them all?


BTW thanks a bunch for adding this PR! I almost was about to try rolling ryot but I'd much prefer to stay here

PR #893, opened 2025-09-29
PWA capabilities

66Bunz:
I noticed that right now it's possible to install the app as a sort of PWA, but it opens pages in new tabs inside the app, so I added real PWA capabilities to Yamtrack.

Let me know if there is something to change, add or remove

Solves #918


66Bunz:
@FuzzyGrim any ETA on getting this merged?

src/app/views.py
FuzzyGrim, line 1: `sw_path = os.path.join(settings.STATICFILES_DIRS[0], 'js', 'sw.js')`

Should this point to `serviceworker.js` instead?

FuzzyGrim: I'm not very familiar with how PWAs work, just a small thing I noticed.

66Bunz: Yes, true, fixed it

FuzzyGrim:
Looks good, thanks for the PR!

PR #907, opened 2025-10-09
Add support for the official Jellyfin Webhook plugin to integrations.html

### LINKED ISSUE #847: [Feature Request] Use official Jellyfin webhook
Jellyfin has an official webhook addon: https://github.com/jellyfin/jellyfin-plugin-webhook/tree/master/Jellyfin.Plugin.Webhook/Templates

It would be preferable to use this than to install an unofficial addon.

There's the unofficial addon, which works :)

Oridjinn1980:
Updated integrations.html to include a Webhook setup using the official Jellyfin Webhooks plugin implementation. I tried to make it pretty and match the original html style, and I also left the old setup using the unofficial repo just in case there was some reason to keep using it.

Also closes #847


src/templates/users/integrations.html
```diff
@@ -50,10 +50,113 @@ <h3 class="text-base font-medium">API Token</h3>
...
+                        "ParentIndexNumber": {{{SeasonNumber}}},
+                        "IndexNumber": {{{EpisodeNumber}}},
+                      {{/if_equals}}
+                      "ProviderIds": {
+                        "Tmdb": "{{{Provider_tmdb}}}",
+                        "Imdb": "{{{Provider_imdb}}}",
+                        "Tvdb": "{{{Provider_tvdb}}}"
+                    },
+                    "UserData": {
+                      {{#if_equals PlayedToCompletion 'True'}}
```
Lucki, line 142: I'm not entirely sure if that matters, but the documentation defines this value as `true/false` (lowercase).

> - Playback
>     - PlayedToCompletion
>         - `true/false`, Only when `NotificationType == PlaybackStop`

FuzzyGrim: It doesn't seem to matter.

Issue #940, opened 2025-10-29
[Feature Request] Ability to Scrobble from Jellyfin


JamsRepos:
**Describe the solution you'd like**  
As the webhook plugin supports scrobbling, progress, pause and resume, we should be able to see progress of watch history within the application.

**Describe alternatives you've considered**  
N/A

**Additional context**  
N/A

PR #941, opened 2025-10-29
feat(progress): added ability to track progress to movies, shows and anime #940

JamsRepos:
Adds Progress Information to Movies, Shows & Anime.
image
image

Solves #940


dannyvfilms:
I'm interested in integrating this into my fork, but I'll need some time to get my head around it to make sure I don't make any mistakes adding it.

FuzzyGrim:
I like the idea of this PR. Tracking partial playback progress for movies, shows, and anime would be a really useful improvement. I think there are three blockers before merge:

1. Plex webhooks will raise `NotImplementedError` because the base handlers now call `_extract_position_and_runtime()`, but `PlexWebhookProcessor` does not implement it.

2. Partial TV progress creates `Episode` rows with `end_date=None`. Existing logic treats any `Episode` row as watched, so a partial progress event on a season finale can mark the season/show completed incorrectly.

3. The existing `progress`/`max_progress` fields represent content units, not playback percentage: TV uses episode count, comics use issue count, and movies use `1` because there is only one movie watch. Changing movies to `max_progress=100` mixes percentage progress into that model and affects event/progress logic. It also means existing data/importers that use `progress=1` for completed movies may display as `1%` unless this is handled with a migration/backfill and importer updates.

66Bunz:
> 2\. Partial TV progress creates `Episode` rows with `end_date=None`. Existing logic treats any `Episode` row as watched, so a partial progress event on a season finale can mark the season/show completed incorrectly.

I think the `Episode` model should be implemented as a Media item. That would simplify a lot of the logic, treating every media type the same.

sboddy:
Author of #1506 - Kodi Integration here. The Kodi Addon I'm piggybacking on has a nice interval event type that passes progress information for played videos every X seconds.
```
            "duration": 2172,
            "progress": {
                "time": 1123,
                "percent": 51.703499079
            },
```
It would be trivial to add that if this PR gets merged at some point.

Following on from @66Bunz I also raised a Q&A because I also tripped over, and was puzzled by, the different handling of episodes.

PR #968, opened 2025-11-12
Allow importing TV seasons/episodes by title

dpantel:
Currently, when importing a TV show (or a movie) via a Yamtrack-compatible CSV file, 'media_id' is optional. If it is not provided, TMDB will be searched for the show Title.

However, the same is not true when importing TV seasons and episodes. If 'media_id' is not provided, the import will result in an error.

This small change allows specifying a show title, without specifying a media_id when importing seasons and episodes.


Dr-Electron:
Technically, parent_type could also be used, as it already is set to TV if it is a SEASON or EPISODE. Maybe that make it more clear or "keeps the changes in one place".

@FuzzyGrim can we get this merged. LGTM

FuzzyGrim:
Thanks!

FuzzyGrim:
Added in `v0.25.0`.

PR #1002, opened 2025-11-29
Hide media card hover buttons on touch devices

andrebk:
## Description of changes

When hovering over the image on media cards, three buttons are shown for tracking, lists, and history. This works great on desktop, but on mobile there is no hover state so they aren't shown. Worse, when tapping the image where the invisible buttons are, you will press the button instead of navigating to the media details page as intended.

This PR hides the entire overlay and the buttons on touch devices (`pointer-coarse`), and adds a link to the `img` itself in addition to the overlay so you can still navigate to the details page when the overlay is hidden. This removes the accidental button clicks on mobile.

### DIFF
--- a/src/app/tests/test_integration.py
+++ b/src/app/tests/test_integration.py
@@ -44,9 +44,9 @@ class IntegrationTest(StaticLiveServerTestCase):
         self.page.get_by_placeholder("Search tv shows...").fill("breaking bad")
         self.page.get_by_role("button").nth(1).click()
         expect(self.page.locator("h2")).to_contain_text("Search Results")
-        self.page.get_by_role("link", name="Breaking Bad", exact=True).click()
+        self.page.get_by_title("Breaking Bad", exact=True).click()
         expect(self.page.get_by_role("main")).to_contain_text("Breaking Bad")
-        self.page.get_by_role("link", name="Season 1").click()
+        self.page.get_by_title("Season 1").click()
         expect(self.page.get_by_role("main")).to_contain_text("Season 1")
         self.page.locator(".p-2").first.click()
         expect(self.page.get_by_role("main")).to_contain_text("Track Episode")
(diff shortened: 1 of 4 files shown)


FuzzyGrim:
> This works great on desktop, but on mobile there is no hover state so they aren't shown. Worse, when tapping the image where the invisible buttons are, you will press the button instead of navigating to the media details page as intended.

This seems like a bug, it should look like this:

https://github.com/user-attachments/assets/35843a6e-c365-4489-a63a-eab4ba1d5db5

This is on iOS, first tap shows the overlay and then you can choose to tap the buttons or the background image to navigate to the details page.

andrebk:
> This is on iOS, first tap shows the overlay and then you can choose to tap the buttons or the background image to navigate to the details page.

On android using firefox, that first tap causes navigation to the details page, so there is no time to press any of the buttons. If you then use the back button to return to the page, the overlay and buttons are showing, which is weird, but more in the way than useful.

https://github.com/user-attachments/assets/7fcb1b1a-8426-4552-9f5f-e08b84a13eb5



On iOS, do you need two presses to click the button, or can you still click it directly while invisible without activating the overlay first?

Even if what I'm seeing is not intended behaviour, I think I would still prefer no overlay on mobile. I want the big tappable target (the image) to navigate me, so I don't have to tap the tiny title. And the overlay isn't as much of a shortcut even with your behaviour. It takes two "clicks" to track something both with the overlay (tap to show, tap to open tracking menu) and with the details page (tap to navigate to details page, tap to open tracking menu). On desktop it's more of a shortcut since hovering is "free".

FuzzyGrim:
Yes, it still requires two clicks, but the big advantage is that you can stay on the same page without having to go to the details page to edit.

But I can see how some people might prefer your method, I'll merge this and then focus on adding a UI preferences setting to allow both options.

PR #1003, opened 2025-11-29
Display movie collections on media detail page

andrebk:
## Description of changes

TMDB has the concept of "collections", which contains movies that are prequels/sequels to eachother. This PR checks a movies metadata if it belongs to a collection, and if it does displays the movies in the collection on the media detail page.

- Hooks into the existing "related" framework as an extra category, so minimal changes are needed in the templates
- Movies in the collection are removed from the recommendations to avoid duplicates
- The current movie in the collection is highlighted

image

## Notes
The collection data can't be fetched using `append_to_response`, since the main metadata must be fetched first to know if there is a collection at all. So for movies with a collection there will be two API requests.

The check for which movie in a collection is the currently viewed movie introduced an if-else statement for the `media_card` component. This is because the active ring around the current movie is added when the media ID matches the collection item ID. But TV seasons are also rendered in this related items loop, and a TV seasons ID is the same as the TV show ID, so all seasons would get the active ring even though they are not part of a collection at all. I didn't see an easy way to fix this without separating out either seasons or collections from other related items, so I went for an if-else instead even though it isn't super DRY.

Collections usually only contain direct sequels/prequels, so for example there is an "Iron Man" collection but no MCU collection, and the "Star Wars" collection contains only the Skywalker saga and not movies like "Rogue One" and "Solo". If you want to check out how these changes look with a large collection, look at any James Bond movie, as they are all part of the same collection (25ish movies).

### DIFF
--- a/src/app/providers/tmdb.py
+++ b/src/app/providers/tmdb.py
@@ -141,8 +141,26 @@ def movie(media_id):
                 url,
                 params=params,
             )
+
+            if response.get("belongs_to_collection", {}) is not None and (
+            collection_id := response.get("belongs_to_collection", {}).get("id")):
+                collection_response = services.api_request(
+                    Sources.TMDB.value,
+                    "GET",
+                    f"{base_url}/collection/{collection_id}",
+                    params={**base_params},
+                )
+            else:
+                collection_response = {}
         except requests.exceptions.HTTPError as error:
             handle_error(error)
+
+        # Filter out collection items from recommendations, to avoid duplicates
+        collection_items = get_collection(collection_response)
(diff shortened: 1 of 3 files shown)


andrebk:
@FuzzyGrim Have you had time to take a look at this at all?

Please don't hestitate to ask me to change or fix things, or even just close the PR if you don't think it's a good fit. I want to help out :smiley:

FuzzyGrim:
Thanks! Looks good.

andrebk:
Thanks @FuzzyGrim 

I saw you made some minor tweaks after merging. In any future PRs, don't hestitate to ask me to make any changes you would like, I'd be happy to do it!

PR #1017, opened 2025-12-04
Hide completed related media if HIDE_COMPLETED is set

### LINKED ISSUE #964: [Feature Request]  Hide Completed movie from Recommendations
Add a dotenv to allow hide a movie **Completed** from **Recommendations** in the movie detail

busliggabor:
Added HIDE_COMPLETED to the environment variables in order to hide related completed media in the detail page.
This resolves #964. I didn't filtered it movies, I think it makes sense in all media types (except seasons, i filtered those out).

Question: how can I add this info to the project's wiki? Or will you be adding this info?

### DIFF
--- a/src/app/helpers.py
+++ b/src/app/helpers.py
@@ -1,6 +1,7 @@
 from urllib.parse import parse_qsl, urlencode, urlparse
 
 from django.apps import apps
+from django.conf import settings
 from django.contrib import messages
 from django.db.models import Q
 from django.http import HttpResponseRedirect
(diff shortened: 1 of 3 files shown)


FuzzyGrim:
Thanks for the PR! I like this feature.

Regarding the wiki: unfortunately, it can't be edited via pull requests, so I'll add the documentation for new environment variables myself.

That said, I'm planning to move these types of user preferences into the new preferences section in the UI rather than keeping them as environment variables. I'll merge this as-is and then make that small change afterwards.

FuzzyGrim:
Released in `v0.25.1`

Issue #1036, opened 2025-12-13
[BUG] Searching for games via IGDB after access token refresh returns zero results


busliggabor:
**Describe the bug**
If the access token is expired/invalid while we search for some game via IGDB, then we get zero results in the retried api call because of bad query parameters.

**To Reproduce**

- Search for a game when the IGDB access token is already invalid (for example: because of expiration), or edit the access token while trying the search in order to get an unauthorized response.
- The retried api call sends empty data, so we get back zero results
- Also because this zero result list is cached, now we don't have valid results until the cache is expired or until we delete it manually

**Expected behavior**
The retried api call should be called with the correct parameters and get the results

**Error logs**
2025-12-13 23:19:52.744
172.20.0.1 - - [13/Dec/2025:23:19:52 +0100] "GET /search?q=Tomb+Raider&media_type=game HTTP/1.1" 200 5548 "https://yamtrack.[redacted]/medialist/game" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36"

2025-12-13 23:19:51.562
warn
[2025-12-13 23:19:51 +0100] [57896] [WARNING] Internet Game Database: Invalid access token, refreshing

**Yamtrack version**: dev
**Database**: SQLite (default)

FuzzyGrim:
Released in `v0.25.1`

PR #1037, opened 2025-12-13
Fix igdb search retry after getting new access token

busliggabor:
This PR fixes the following ticket: #1036


FuzzyGrim:
Nice catch!

FuzzyGrim:
Released in `v0.25.1`

PR #1044, opened 2025-12-16
Feature: Add "Update" mode to Steam Importer

Laudenlaruto:
This PR introduces a new "Update" mode for the Steam importer, allowing users to sync progress (playtime) and status for existing games without overwriting custom metadata or resetting completed games.

Following up on PR #711 

Key Changes:

- Backend (steam.py):
  - Implemented _update_existing_game to handle syncing progress and status.
  - Updated _process_game to support the new update mode.
  - Refined new mode to strictly skip existing items (previously it might have processed them depending on flow).
  - Added logic to protect "Completed" or "Dropped" statuses from being overwritten by the importer, while still updating playtime.
 
- Frontend (import_data.html):
  - Added "Sync New Items and Update Existing" (value="update") option to the import settings dropdown.
  
- Testing (test_steam_update.py):
  - Added comprehensive tests covering:
    - Updating progress and status for "Planning" or "In Progress" games.
    - Ensuring "Completed" games update playtime but retain their status.
    - Verifying that "New" mode ignores existing games entirely.

How to Test:

  - Navigate to the Import page
  - Select "Sync New Items and Update Existing" from the Import Mode dropdown.
  - Run a Steam import.
  - Verify that:
    - New games are added.
    - Existing games have their playtime updated.
    - Existing
  
  Games marked as "Completed" do not revert to "In Progress" even if playtime increased.

My concern:

- Introducing the "update" mode will possibly mislead or even break import from other provider (Trakt, AniList, MAL, IMDB, etc.) There are a couple of ways to tackle this : 
  - Either update the `should_process_media` to default back to new with a warning for source that don't support this method
  - Implement "Update" logic for all other importers (Trakt, AniList, etc.) to handle the new mode.

WDYT ?


FuzzyGrim:
Thanks! About the concern, I think instead of adding new "Update" mode, I will replace what "Overwrite" does with your changes.

Issue #1051, opened 2025-12-21
[BUG] Hardcover API field 'recommendations' not found in type


artifaxx:
**Describe the bug**
Opening a book returns a 500 error.

**To Reproduce**
Open any book.

**Expected behavior**
The book details page opens.

**Error logs**
```
[2025-12-21 09:33:46 -0800] [209] [ERROR] Internal Server Error: /details/hardcover/book/44503/the-woman-in-the-dunes
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django/views/decorators/http.py", line 64, in inner
    return func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/yamtrack/app/views.py", line 197, in media_details
    media_metadata = services.get_media_metadata(media_type, media_id, source)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/yamtrack/app/providers/services.py", line 211, in get_media_metadata
    return metadata_retrievers[media_type]()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/yamtrack/app/providers/services.py", line 206, in <lambda>
    MediaTypes.BOOK.value: lambda: hardcover.book(media_id)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/yamtrack/app/providers/hardcover.py", line 159, in book
    book_data = response["data"]["books_by_pk"]
                ~~~~~~~~^^^^^^^^
KeyError: 'data'
```

```
<html><body>
<!--StartFragment-->
[...]

PR #1052, opened 2025-12-21
Removed deprecated recommendations  endpoint calls from Hardcover pro…

busliggabor:
Fixes the following issue: #1051
Removed the recommended part from Hardcover provider, along with the unit tests belonging to that part of the code


FuzzyGrim:
Thanks for the PR!

PR #1059, opened 2025-12-26
Fix duplicate width/height attributes for SVG Icons

xVaDiMx:
All SVG Images currently have width and height set as:
```
width="auto"
height="100%"
```
However, when these SVG images are included, a class with width and height `classes="w-5 h-5"` is added again, causing conflicting CSS rules. This results in errors in the browser console.

**Changes:**
Removed the redundant width/height classes to prevent conflicts and fix the console errors.

Screenshot_20251226_200056


FuzzyGrim:
Nice, thanks!

FuzzyGrim:
Released in `v0.25.1`

PR #1067, opened 2025-12-31
Add obfuscate unseen episodes user setting and functionality

corymills:
# Summary
Adds a new user setting that allows users to hide episode artwork and blur episode descriptions for unseen episodes, helping avoid accidental spoilers.

# Changes
- New User Setting: `obfuscate_unseen_episodes` field in the User model (defaults to off)
- UI Preferences Page: Added toggle in the UI Preferences settings to enable/disable the feature
- Media Details Template: Episode images are replaced with media poster artwork and descriptions are blurred when the setting is enabled and the episode hasn't been watched
- Database Migration: Added migration for the new user field

image

image

# Testing
- Manual testing of the feature
- Added unit tests


66Bunz:
It could be nice if when clicking the hidden image/plot it reveals

corymills:
> It could be nice if when clicking the hidden image/plot it reveals

I like that idea a lot.  I've added this to the PR.

andrebk:
Would it be better to blur the episode image, rather than using the season image in its place? Considering they usually use different aspect ratios, I would expect a lot of weird clipping when used like this.

FuzzyGrim:
Thanks for the PR! 

I added this: 

> Would it be better to blur the episode image, rather than using the season image in its place

amadejpapez:
Should episode title also be blurred? That's how Trakt did it - it showed only "Episode X" which is here already shown below.

PR #1093, opened 2026-01-11
Enhance docker-compose security settings

pstoeckle:
Hi there,

I really like the repo, and host a copy in my local lab. Thanks for developing it! I added some docker-compsoe security measures there, maybe you are interested in them:

- Add only need linux capabilities to containers
- Readonly filesystems for postgres/redis
- Network layout: web available from outside, on web can talk to redis, only web can talk to db
- Add healthchecks

### DIFF
--- a/docker-compose.postgres.yml
+++ b/docker-compose.postgres.yml
@@ -4,8 +4,8 @@ services:
     image: ghcr.io/fuzzygrim/yamtrack
     restart: unless-stopped
     depends_on:
-      - db
-      - redis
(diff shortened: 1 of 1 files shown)


FuzzyGrim:
Thanks for the PR and I'm glad you're enjoying the project! I appreciate the security hardening, but I want to keep the default compose files as minimal as possible to ensure they are easy to read and debug. Users with specific security requirements can easily customize it to their needs.

PR #1097, opened 2026-01-12
Hide media score if rating is 0

busliggabor:
Sometimes the media items have 0 ratings, which looks strange in my opinion. But it could be used as a feature as well, for example if someone wants to rate something later, then it can be used to indicate that intent. So, in order to not break that workflow, I created an environment variable to hide the 0 rating from the UI.
Usage: set HIDE_ZERO_RATING environment variable to 'True'


FuzzyGrim:
Thanks!

Issue #1071, opened 2026-01-02
[BUG] Plex scrobbler randomly fails


Floydian-dk:
**Describe the bug**
When playing tv-episodes from non-English countries in Plex, scrobbling fails and plays are not recorded in Yamtrack. 

**To Reproduce**
Play an episode from a non-English TV-show, and observe that nothing is recorded in Yamtrack.
Have a notion that this may happen because I have set up my library settings in Plex to "Use original titles", and not use the default language of the interface.

**Expected behavior**
Playback of episode should be recorded in Yamtrack

**Screenshots**
n/a

**Error logs**
Episode of Spanish TV show Aliados fails:
[2026-01-02 22:19:52 +0100] [85713] [INFO] Extracted IDs from payload: {'tmdb_id': '1611389', 'imdb_id': None, 'tvdb_id': None}
[2026-01-02 22:19:52 +0100] [85713] [INFO] Received webhook for tv: Aliados S01E01
[2026-01-02 22:19:52 +0100] [85713] [WARNING] No matching TMDB ID found for TV show

Episode of Danish TV show "Bedrag" fails:
[2026-01-02 22:06:52 +0100] [85713] [INFO] Extracted IDs from payload: {'tmdb_id': '1148341', 'imdb_id': None, 'tvdb_id': None}
[2026-01-02 22:06:52 +0100] [85713] [INFO] Received webhook for tv: Bedrag S01E01
[2026-01-02 22:06:52 +0100] [85713] [WARNING] No matching TMDB ID found for TV show

Episode of US TV-show "Heroes" is recorded correctly:
[2026-01-02 22:07:27 +0100] [85713] [INFO] Extracted IDs from payload: {'tmdb_id': '84128', 'imdb_id': 'tt1041277', 'tvdb_id': '332474'}
[2026-01-02 22:07:27 +0100] [85713] [INFO] Received webhook for tv: Heroes S02E01
[2026-01-02 22:07:27 +0100] [85713] [INFO] Detected TV episode via TMDB ID: 1639, Season: 2, Episode: 1

**Yamtrack version**:  v0.24.11
**Database**: PostgreSQL

Floydian-dk:
Have done some further testing, and this issue is not specifically related to the original language of the TV-show as I assumed above.
Plex scrobble also fails for English titles, so this appears to be more of a random issue, than tied to something specific.

When matching fails I see different warnings in the log:
[...]

PR #1109, opened 2026-01-17
fix: [BUG] Plex scrobbler randomly fails #1071

Floydian-dk:
Attempt at fixing the issue with Plex scrobbling of TV-episodes when only TMDB ID is provided in the Plex payload.

TMDB does not allow to look up TV-episodes by their ID, so a workaround is needed when TMDB does not have either IMDB ID or TVDB ID for a given episode.

The workaround below is to search for TV-shows by their name, then check for each returned TV-show if we can match an episode by season nr, and episode nr. and find a match for the TMDB episode id provided in the Plex payload.

In tmdb.py an additional parameter (ID) has been added to the result set.
In base.py an initial check is done to determine if we have either IMDB or TVDB ID for the episode, if not, the mentioned search willl be done, else the previous lookup is used.


src/integrations/webhooks/base.py
```diff
@@ -97,43 +98,69 @@ def _process_tv(self, payload, user, ids):
...
-        )
         if not media_id:
             logger.warning(
                 "No matching TMDB ID found for TVDB episode ID: %s", tvdb_episode_id
             )
-            return
 
+            if ids["tmdb_id"]:
+                # We have tmdb_id for this TV-show episode.
+                series_name = payload["Metadata"].get("grandparentTitle")
```
FuzzyGrim, line 138: This file is the base file for all webhooks, here we are handling some Plex specific payload shape.

src/integrations/webhooks/base.py
```diff
@@ -97,43 +98,69 @@ def _process_tv(self, payload, user, ids):
...
+
+                response = app.providers.tmdb.search(MediaTypes.TV.value, series_name, 1)
+
+                for result in response["results"]:
+                    show_id = result["media_id"]
+                    try:
+                        episode_data = app.providers.tmdb.episode(show_id, season_number, episode_number)
+                        if episode_data["tmdb_id"] == int(ids["tmdb_id"]):
+                            logger.info("We have matched tmdb_id using search")
+                            media_id = result["media_id"]
```
FuzzyGrim, line 153: Could you add tests for this new logic?

src/app/providers/tmdb.py
```diff
@@ -713,6 +713,7 @@ def process_episodes(season_metadata, episodes_in_db):
                 "episode_number": episode_number,
                 "air_date": episode["air_date"],  # when unknown, response returns null
                 "image": get_image_url(episode["still_path"]),
+                "tmdb_id": episode["id"],
```
FuzzyGrim, line 716: Because this changes the metadata returned by the TMDB provider by adding tmdb_id, please bump the Django cache VERSION in settings.py. We do this for provider metadata shape changes so users do not keep stale cached metadata after upgrade.

Dockerfile
FuzzyGrim, line 1: This file should not be in this PR

FuzzyGrim (changes requested):
Thanks for working on this. I added a few inline comments.

I’m also not fully sure about the overall lookup strategy here. Besides the cost of iterating through search results, TMDB episode IDs are not especially stable, which is likely part of why TMDB does not support direct episode lookup by `episode_id`. 

Maybe the bigger question is why those episodes do not have IMDb or TVDB IDs in the first place.

Floydian-dk:
Thank you for your feedback on my contribution.

Since making this pull request, I've been working a bit more on this issue on my fork, and stumbled upon some logic issues.
Turns out that Plex can return TVDB and IMDb ID's even though they do not exists in TMDB, which causing the logic in this pull request to fail.
I assumed that if Plex provided these ID's, that they would exists in TMDB, but that is not always the case. Based on my observations/test, it appears that Plex is having their own logic to match and merge data from IMDB and TVDB, with their TMDB feed, and this is the included in the webhook data. 

So I've improved the logic, to first try to use TVDB ID, and if that fails, then fall back to the matching using TMDB show ID and season/episode number, implemented above.

This so far works for me... If it breaks stuff elsewhere, I can't tell.

When implementing my fix for the 0.25.2 version, it appeared to me that the lookup by IMDb ID has been removed as part of the handling of Anime which was implemented in this release?

As for why TVDB &/or IMDB  ID's are missing in TMDB, there can be several reasons.
I've been contributing quite a lot to TMDB, and for some of the shows I follow, there may not be an IMDB entry at the time when I submit data to TMDB. As for TVDB ID's I very seldom check for those. When Plex switched to using TMDB, I didn't see the need. Within a day or so after adding an episode to TMDB, Plex would have it, and be able to match the episode in my library, so no need to check for TVDB data.

Floydian-dk:
This is now obsolete.

A possible fix using newer code as been added in this pull-request: https://github.com/FuzzyGrim/Yamtrack/pull/1627

PR #1128, opened 2026-01-25
rootless container

smerschjohann:
This PR allows to run the container rootless and does not change any container files during runtime


FuzzyGrim:
This looks like a regression for bind-mounted volumes.

The container now always runs as UID/GID 1000, and the runtime PUID/PGID remapping was removed. 

Also, on first boot `python manage.py migrate` will fail with a permission error.

smerschjohann:
yes. If you want to run it as a different id, you should use the native docker / podman / kubernetes methods, instead of the "I need root to downgrade to a user id" way.

PR #1130, opened 2026-01-26
[Feature] Progress line on media items

busliggabor:
This PR adds the ability to display a progress bar on tracked media items. This behaviour can be toggled in the settings via the "Show progress bar on media items" checkbox.

The progress bar colours are the following:
- Green: Completed
- Blue: In progress
If an item doesn't have max progress (for example: games), then the line is dotted
- Purple: In progress, all existing episodes/progress caught up
- Yellow: Paused
- Red: Dropped
- White: Planning

Images:
Home page
home_page
Setting page
Settings
Anime page
anime


66Bunz:
I personally think better colors can be chosen to be more in line with Yamtrack current UI...
Something custom rather than normal red, green ecc colors.

busliggabor:
Maybe the exact colours that the status icons already have? My only gripe with that is that the "in progress" and "planning" colours are very similar now on the current UI in my opinion.
image
image
Also, I am thinking about adding an environment variable/database entry to setting the colours, so it could be set to whatever the user want it to be.

66Bunz:
> Maybe the exact colours that the status icons already have? My only gripe with that is that the "in progress" and "planning" colours are very similar now on the current UI in my opinion.

That's true, but maybe having a full bar it shows the differences better? Or maybe the two colours could be slightly adjusted to drive them more away from each others?

> Also, I am thinking about adding an environment variable/database entry to setting the colours, so it could be set to whatever the user want it to be.

I think that if/when Yamtrack will have themes it'll be the right way, but since it doesn't yet, the colors shouldn't be modifiable.

busliggabor:
I changed the colours based on the ones already in the app. The only change I am not sure I can make without discussing is the following: the colour of the "Planning" status is currently blue, but it is too similar to the "In progress" indigo colour. I think changing bluet to cyan would make it look better. It would change the colour of the icon as well, so it would stay consistent across the app (I found the status colours in config.py STATUS_CONFIG. Here is two image comparing them:
With the original blue colour:
recolored_blue
With the changed cyan colour:
recolored_cyan
Also I removed the dotted style for items without max progress, and the "In progress" status uses only one colour, regardless the fact that it is all caught up in episode number or not.
What do you think about it @66Bunz  @FuzzyGrim ? I haven't uploaded the new commit yet, but will be uploading if I got feedback about it.

66Bunz:
For me the second option is better, but I'd wait for fuzzygrim's response

FuzzyGrim:
I think cyan stands out a bit too much, maybe "sky" could be a good middle ground?

But the screenshot looks awesome in general!

busliggabor:
Here is with sky colour:
recolored_sky
I used the colour pallette from here [https://tailkit.com/tools/tailwind-color-palette](https://tailkit.com/tools/tailwind-color-palette)
I will upload the new commit shortly.

src/app/views.py
```diff
@@ -45,6 +45,7 @@ def home(request):
     if request.headers.get("HX-Request") and media_type_to_load:
         context = {
             "media_list": list_by_type.get(media_type_to_load, []),
+            "user": request.user,
```
FuzzyGrim: You don't need to add `user` to the context, it's already available in the template with `request.user`.

busliggabor: Removed it.

FuzzyGrim:
Added in `v0.25.0`

PR #1146, opened 2026-02-02
Support providing playtime as float

KucharczykL:
Currently the playtime can be provided with a single number of hours as integer but I find myself often only knowing the total playtime as float (e.g. Steam only provides this number in the app's UI), for example 3.5 to signify 3 hours 30 minutes.

Supporting this float notation is very easy and works well with the established system. I have also changed the error message to include more information on what's allowed though it's getting a bit unwieldy.


FuzzyGrim:
Thanks @KucharczykL!

PR #1148, opened 2026-02-03
Add maskable icons in webmanifest for PWA

pogman-code:
Add maskable icons for the progressive web app.

I guess a before/after comparison would be better to explain the purpose of this change, so there it is:

Before:
image

After:
image

for more explanation, see: https://web.dev/articles/maskable-icon and https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Define_app_icons#support_masking


pogman-code:
I've made a second commit to add shortcuts icons:
image

image

FuzzyGrim:
Thanks @pogman-code! Looks great!

PR #1202, opened 2026-02-15
Adding dropdown options for episode watch

McBaumwolle:
As I am migrating from trakt (like many others) I missed the "check-in" feature, which automatically marks an episode (or film) as watched when its runtime is over. This PR implements something similar by adding 3 drop down options to the "Add watch" for episodes (only!). One fills in the current date and time, the other fills in the current date and time plus the runtime of the selected episode. And the third is the one which inserts the "air date" for consistency.

<details>
  <summary>Screenshot</summary>

image

</details>

<details>
  <summary>Screencast</summary>

The clear button glitch is only in the compressed video.

https://github.com/user-attachments/assets/d32b4bfd-5224-46d6-81a7-9b11dda73a30

</details>

First I tried it with a button left to the "Add watch", but I feel this looks cleaner. I would create a second PR for the same functionality for movies, but I did not want to make too many changes at once. ^^

(I also made some background changes to the episode history section in the popup so that the new button works better.)


McBaumwolle:
To stay consistent, I also moved the "air date" button to the drop down and added a clear button.

FuzzyGrim:
Thanks for your contribution! It looks great!

I decided to remove the delete button, as I believe the browser's built-in clear button should be clear.

PR #1203, opened 2026-02-15
Show watch providers on media details page

### LINKED ISSUE #420: [Feature Request] Provide Watch Provider/Streaming Provider information
When browsing Planned movies/TV shows it would be helpful to be able to know what streaming service the user subscribes to the movie/show is available on.

This data is available from TMDB for Movies and TV shows at least. It may require the user to specify the country they're interested in streaming services from (some movies have very long lists otherwise -- say Disney+ in every region). The user may potentially also want to indicate the services they have access to.

Providing the service information on the detail screen would be great.

(Even better would be also providing it on list screens but these would need additional work first so you could pick fields to be shown on the list screens)

andrebk:
## Changes
Show watch providers (streaming services) on media detail page. This closes #420.

- User can select which region they belong to in the preferences, to get correct providers
- If no region is selected, the section will not be shown at all to support users with their own libraries on plex/jellyfin etc.
- Free and flatrate (subscription) providers are shown, rental and purchase platforms are available in the API but not shown
- If a region is selected and no providers are available, "No watch providers" will be shown

## Screenshots

<details><summary>Placement on media page</summary>
image
</details>

<details><summary>Many providers</summary>
image
</details>

<details><summary>No providers</summary>
image
</details>

<details><summary>Region selection</summary>
image
</details>

## Notes

- I don't know what icon source is used for Yamtrack, so i grabbed a globe icon from [pictogrammers](https://pictogrammers.com/) for the settings. Let me know if this should be changed to match the other icons.
- I decided to exclude rental and purchase platforms as it was difficult to include while keeping the design compact, since they need to be separated from streaming platforms. I don't rent/buy media, so for me this feels OK, but let me know if it should be included.


FuzzyGrim:
Nice!

I think not including rental and purchase platforms is fine.

The icons I use are from [lucide](https://lucide.dev/). Does the TMDB API return links to the different platforms?

andrebk:
> Does the TMDB API return links to the different platforms?

Yes and no. They don't provide any links to each streaming platform, but they do provide a link to a TMDB page that does. I assume they get some kickbacks or something when their links are clicked. From the [TMDB docs](https://developer.themoviedb.org/reference/movie-watch-providers):

>This is not going to return full deep links, but rather, it's just enough information to display what's available where.
>
>You can link to the provided TMDB URL to help support TMDB and provide the actual deep links to the content.

Should I add it? Maybe clicking "streaming"?

> The icons I use are from [lucide](https://lucide.dev/).

Nice! They have a globe icon as well, I'll update it.

What do you think of the gemini comments? 2 and 3 look like easy fixes, the other two are a little more work.

src/app/providers/tmdb.py
```diff
@@ -724,3 +759,37 @@ def episode(media_id, season_number, episode_number):
...
+                "GET",
+                url,
+                params=params,
+            )
+        except requests.exceptions.HTTPError as error:
+            handle_error(error)
+
+        data = [("", "No Region")]
+        regions = response.get("results", [])
+        for region in sorted(regions, key=lambda r: r.get("english_name", "")):
```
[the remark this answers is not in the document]
FuzzyGrim: TMDB API doesn't seem to return null

src/app/providers/tmdb.py
```diff
@@ -274,14 +275,19 @@ def enrich_season_with_tv_data(season_data, tv_data, media_id, season_number):
...
 
     for i in range(0, len(season_numbers), max_seasons_per_request):
         season_subset = season_numbers[i : i + max_seasons_per_request]
-        append_text = ",".join([f"season/{season}" for season in season_subset])
+        append_text = ",".join(
+            [
+                f"season/{season},season/{season}/watch/providers"
+                for season in season_subset
+            ]
+        )
```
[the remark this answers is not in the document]
FuzzyGrim: This is false.

src/app/providers/tmdb.py
```diff
@@ -434,10 +443,11 @@ def process_tv(response):
         "external_links": get_external_links(response.get("external_ids", {})),
         "last_episode_season": last_episode["season_number"] if last_episode else None,
         "next_episode_season": next_episode["season_number"] if next_episode else None,
+        "providers": response.get("watch/providers", {}).get("results", {}),
     }
 
 
-def process_season(response):
+def process_season(response, providers_response=None):
```
FuzzyGrim: When should `providers_response` not be passed? I don't like using optional arguments for no reason.

src/users/views.py
```diff
@@ -251,6 +253,7 @@ def preferences(request):
         TimeFormatChoices.HOUR_24,
     )
     media_types_checked = request.POST.getlist("media_types_checkboxes")
+    request.user.watch_provider_region = request.POST.get("watch_provider_region", "")
```
FuzzyGrim: We should check if the value is valid.

FuzzyGrim (changes requested):

src/app/providers/tmdb.py
andrebk: I made it optional for backwards compatibility, but you're right its always provided, so made it mandatory

src/users/views.py
andrebk: Added, it will now check if the preferred region is in the list of available regions before storing. If region is not valid, `no region` option will be set instead.

FuzzyGrim:
Thanks for the changes!

Currently, the default region is `""`, and with this option, it seems the user won't see any watch providers and therefore, might miss this feature.

I believe we should provide a default value in `users/models.py`. And, if the selected option in the preferences page isn't correct, we can ignore it and revert to the previous one.

andrebk:
You're right, there's a chance the feature is missed. I think there are three cases that we must be able to configure:

1. A region is set, watch providers are shown (normal case)
2. No region is set, show text on media page encouraging user to set a region (initial/default case)
3. Watch provider feature is explicitly disabled by user, nothing is shown on media page

Currently `""` is used both for "no region" and "explicitly disabled", so we can't differentiate between them and show something to the user initially so they can discover the feature.

I suggest keeping `""` as the value for "explicitly disabled", but adding a special `UNSET` string for the "no region" case. `UNSET` will be the default in the model, and is what will be set if the POST request contains an invalid region so that the user sees the feature is misconfigured instead of it disappearing from the media page. I'm thinking `UNSET` should not be in the preferences dropdown though, since after feature discovery the user will want to either set a region or disable the feature, so no need to be able to set `UNSET` explicitly.

Thoughts @FuzzyGrim ?

FuzzyGrim:
Right! I forgot about the disabled by user option. But I wonder if we really need the "no region set" state, we could have the default, for example, `US`, with a tooltip about the ability to change the region in the preferences. That way, we wouldn't need to handle the second case.

Then if selected region is invalid, we could ignore it and not save this option.

And have the disabled option in the preference dropdown.

andrebk:
I don't like the idea of defaulting to any particular region. That will mean wrong provider data is shown to all users who are not from that region. 

Yes, you could have a tooltip to tell you it's possible to change the region, but then that tooltip would have to always be there, as there would be no way to tell the difference between someone who has the US region because they set it explicitly, or just because it is the default. Seems overkill for a setting you would probably only need to set once. 

I think this will end up with users confused why the provider data they're shown doesn't make sense for them. It's better to have an explicit "no region set" state, so that the user is never shown wrong data, but instead a message to go to the preferences and set a region.

andrebk:
I made an implementation of my suggestion for your consideration. For new users this is now shown on the media page:

image

**NOTE: I modified the original migration file with the new default, so it might not apply automatically**

As a non-US user, I feel quite strongly that having a US region default is a bad idea, as that is the kind of thing that annoys me in other apps. I feel like showing non-US users wrong data just to make the feature discoverable is more harmful than helpful, there are probably better ways to do discoverability. That being said, this is your project, if that's the way you decide to go I will change this implementation to match.

McBaumwolle:
As another non-EU user I agree, maybe you could also add a link in the text (or separate button) to the setting.

andrebk:
> As another non-EU user I agree, maybe you could also add a link in the text (or separate button) to the setting.

The word "preferences" in the text is a link, it just doesn't look like one until you hover over it, just like e.g. source or external links in the details box.

FuzzyGrim:
Released in `v0.25.1`

PR #1207, opened 2026-02-16
Enhancing the new cast list

McBaumwolle:
This PR (draft as it is still buggy but wanted to put it there anyway for review, etc.) changes the handy new cast view (from @connorjburton) to be horizontally scrollable, similar to what trakt does at the moment. 

<details>
  <summary>Details</summary>
  
Screenshot From 2026-02-16 12-09-10

</details>

I tried a separate button (like discussed [here](https://github.com/FuzzyGrim/Yamtrack/pull/1168#issuecomment-3902747206)) which did not work (for me) and was kinda inconvenient when there was a giant cast list. I increased the cast fetch number from 10 to 30, and if there are more it adds a button/card in the end linking to TMDB. This is a bit buggy right now, hence the draft. 

## Issues
- [x] Calculation bug
- [x] The side bar gets squeezed a bit - Somewhere I broke the sidebar, or the changed cast section pushes it away.


McBaumwolle:
I tried all the suggestions from the bot about Tailwind instead of inline styles, but the layout kept collapsing (the cast cards then only had the width of the person's name). I'd suggest keeping the inline styles, but @FuzzyGrim has to decide on this as it's not a perfect solution. As far as I see it, it does not break parent columns (right now!?) and the rest of the PR works as expected on my end. Juts let me know if that works for you, otherwise I just have to learn some Tailwind haha.

FuzzyGrim:
Thanks! It looks great!

I can try fixing the inline styling as a follow up.

FuzzyGrim:
Released in `v0.25.1`

PR #1208, opened 2026-02-16
feat(add-docs): Add documentation

### LINKED ISSUE #1206: [Feature Request] Add documentation website
Having a documentation website instead of the GitHub Wiki could really improve how users can consume and improve and contribute to the Yamtrack documentation

66Bunz:
This closes #1206

Preview: https://66bunz.github.io/Yamtrack/

This PR adds the documentation (copied from the current Wiki) using [Material for MkDocs](https://squidfunk.github.io/mkdocs-material)

Also the build workflow was added at [build-docs.yml](https://github.com/66Bunz/Yamtrack/blob/696fb7b59dbc62bb8198208f4a935a6bb7c0b626/.github/workflows/build-docs.yml), the action was copied from the [official documentation](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#with-github-actions).

To deploy using GitHub Pages, from the settings of the repo > Pages > Source the `gh-pages` branch needs to be selected.


.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,30 @@
+name: build-docs
+on:
+  push:
+    branches:
+      - main
```
FuzzyGrim: This repo doesn't have the `main` branch

66Bunz: Sorry, fixed now

FuzzyGrim:
I think the documentation should show what we have in the `release` branch so that users don't get confused when they see something in the documentation that isn't available yet.

66Bunz:
> I think the documentation should show what we have in the `release` branch so that users don't get confused when they see something in the documentation that isn't available yet.

I don't agree completely. The documentation should be done at the same time as the features/fixes implemented. The `dev` branch should have the documentation of features not yet in a release.
Yes it can be confusing to see documentation for stuff not already in a release tho.
Versioning can also be added to the documentation, so it's possible to have docs for the `dev` branch for example in a `dev` tag and the documentation of the releases under the same version number.
More here: https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

FuzzyGrim:
Yes, I agree that the documentation should be written concurrently with development. I was referring to change GitHub's action to generate the live documentation only to be triggered when merging `dev` into `release`.

Versioning also seems like a good idea; perhaps it could be implemented as a follow up pull request.

66Bunz:
Ah ok, everything is clear now.

Changed the branch to `release`, I'll make a new PR for the versioning thing in the future.

FuzzyGrim:
I can work on the doc versioning

PR #1228, opened 2026-03-01
Add support for removing items from a list

olioby:
Here's another PR to allow easily removing an item from a list. There aren't any new views, this just adds a button to use the `list_item_toggle` endpoint.

### Screenshots
image

---

Note: including this confirmation is up to you, it's a one line change.
image


olioby:
^ I'm going to defer to you on which ai comments you think should be addressed

PR #1240, opened 2026-03-08
Add user preference for showing adult titles.

### LINKED ISSUE #606: [Feature Request] Add an option to show or hide NSFW results in the user's settings page
If any of the NSFW environment variables are set to `True` then show an option in the user's setting page to show/hide NSFW results, preferably defaulting to `hide` for new user accounts.

Setting the NSFW environment variables to `False` is not desirable since its effect is global.

**Additional context**
The desired outcome of this feature is to have a way for NSFW visibility to be account specific. For those that want to globally disable it they can still do so from the environment variables which will also hide the option in the settings.

gregersn:
Closes #606


src/users/migrations/0057_user_show_adult_titles.py
FuzzyGrim, line 1: This needs a rebase on dev. The migration is numbered 0056 and depends on `0055_user_jellyfin_mark_events`, but dev already has `0056_user_home_hide_unreleased`

src/app/providers/igdb.py
```diff
@@ -192,7 +192,7 @@ def search(query, page):
             f'where name ~ *"{query}"* & game_type = (0,1,2,3,4,5,6,7,8,9,10)'
         )
 
-        if not settings.IGDB_NSFW:
+        if not settings.IGDB_NSFW or not show_adult_titles:
```
FuzzyGrim, line 195: Let's remove the global setting and only have the per account setting.

gregersn: I thought about that as well, but I guessed that there would probably be people who run an instance for their family, or similar, that do not want to let children, etc, choose for themselves. Or if it is in use in institutions or, whatever. 
Just to raise that point.

PR #1244, opened 2026-03-09
Import Watcharr JSON

JodliDev:
This adds an option to import the exported JSON from https://github.com/sbondCo/Watcharr
It only imports the "final" state of each entry. So even though the activity for each entry can be found in the Watcharr JSON, I opted to ignore it for now.

The implementation reuses the basic YamtrackImporter and just reformats the data from JSON into CSV.

I mainly wrote this for myself but extended it as a PR to hopefully be helpful for others too.
Thank you for this great project btw! :)


66Bunz:
Just wanted to ask the format of the Watcharr dates, because I think that importing the whole `python-dateutil` module might be a bit overkill. Maybe the same result could be reached using the default libraries?

JodliDev:
Yes, good point. It should definitely be possible with standard library. My python game is just a bit rusty and I was going safe rather than sorry :)

From what I can see, the format of watcharr is also ISO 8601 but with the Z notation instead of "+00:00" and an undefined number of digits for the fractional seconds (e.g. `2024-07-26T15:16:49.6934216Z` or `2024-08-02T20:22:57.407627351Z`). So simple string manipulation should do the trick.
I will push a commit

JodliDev:
I am still motivated to make changes if needed. If there is no plan to ever include it, I understand. But please let me know :)

PR #1256, opened 2026-03-19
Grammer correction

LAUGHINGCATMEME:
SQlite &rarr; SQLite
a iCalendar &rarr; an iCalendar
self hosted &rarr; self-hosted

ps: My first contribution so it's very small but I'll try to do more in future.


KaiStarkk:
Hacktoberfest is still six months away, but the spirit lives on year-round

PR #1264, opened 2026-03-22
Fix: Set executable permissions for entrypoint.sh during copy

glensc:
Update entrypoint.sh copy command to set executable permissions directly in the Dockerfile.

This avoids `entrypoint.sh` file added twice to docker layers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ARG VERSION=dev
 ENV VERSION=$VERSION
 
 COPY ./requirements.txt /requirements.txt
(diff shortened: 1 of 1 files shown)


glensc:
maintainers have not responded anything in the PR. so, i'm reluctant to resolve conflicts

66Bunz (approved):
Actual space savings are not that relevant, but it's more polished.
About privileges, changing the permissions to `700` doesn't solve any possible exploit I think. Again it's polished.

PR #1282, opened 2026-03-28
Migrate to uv with pyproject.toml to specify build-system and dependencies 

minerop5000:
Reasons explained in https://github.com/FuzzyGrim/Yamtrack/discussions/1050

Changes made:
- Mode requirements.txt to pyproject.toml with UV
- update GitHub workflow
- updated Dockerfile
- update readme

Not done:
- tested CI/CD - please check for yourself
- automation of the version tag in the pyproject.toml 
- upgrade dependencies to newer versions


pyproject.toml
```diff
@@ -22,3 +69,43 @@ indent = 2
 [tool.coverage.run]
 concurrency = ["multiprocessing"]
 omit = ["*/migrations/*", "*/tests/*", "*/__init__.py", "*/admin.py"]
+
+[dependency-groups]
+dev = [
```
jakobkukla, line 73: Dependency groups are additional to the main build deps. You'd only want to add the actuall dev dependencies here. I had something like this in my draft (note: version might differ):

```
[dependency-groups]
dev = [
  { "include-group" = "testing" },
  { "include-group" = "lint" },
]

testing = [
  "coverage~=7.13.0",
  "fakeredis~=2.32.1",
  "pytest-django~=4.11.1",
  "pytest-playwright~=0.7.2",
  "tblib~=3.2.2",
]

lint = [
  "djlint~=1.36.4",
  "pre-commit~=4.5.1",
  "ruff~=0.14.10",
]
```

pyproject.toml
```diff
@@ -1,11 +1,58 @@
+[project]
+name = "yamtrack"
+version = "0.25.0"
+description = "A self hosted media tracker."
+readme = "README.md"
+requires-python = "==3.12.*"
```
jakobkukla: Probably rather `>=3.12`? Not sure though.

pyproject.toml
```diff
@@ -1,11 +1,58 @@
...
+dependencies = [
```
jakobkukla, line 7: `pyproject.toml` is for specifying "abstract" dependencies. Meaning they should specify a range of versions that should work (usually something like `>=1.0` or even better relying on semantic versioning `~=1.0`. Locking these to a "concrete"/specific version for reproducibility is `uv.lock`s job. This has the benefit that uv will take the latest version that's compatible when creating the lockfile.

pyproject.toml
```diff
@@ -1,11 +1,58 @@
...
+
 [tool.ruff]
 target-version = "py312"
 extend-exclude = ["migrations"]
 force-exclude = true
 
 [tool.ruff.lint]
 select = ["ALL"]
-ignore = ["ANN", "PT", "PD", "D100", "D104", "RUF012", "PLR0913", "SLF001", "COM812"]
+ignore = [
```
jakobkukla, line 44: unrelated whitespace change

Dockerfile
```diff
@@ -18,8 +21,7 @@ RUN sed 's/listen 8000;/listen 8000; listen [::]:8000;/' /etc/nginx/nginx.conf >
 WORKDIR /yamtrack
 
 RUN apk add --no-cache nginx shadow \
-    && pip install --no-cache-dir -r /requirements.txt \
-    && pip install --no-cache-dir supervisor==4.3.0 \
+    && uv sync --locked \
```
FuzzyGrim, line 24: How will this work without supervisor?

.github/workflows/app-tests.yml
```diff
@@ -51,19 +51,18 @@ jobs:
...
+        uses: astral-sh/setup-uv@v7
         with:
           python-version: ${{ matrix.python-version }}
 
       - name: Install Dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements-dev.txt
+          uv sync --locked --all-extras --dev
           playwright install
```
FuzzyGrim, line 61: Tests are failing with `playwright: command not found`

pyproject.toml
```diff
@@ -1,11 +1,57 @@
...
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+	"aiohttp~=3.13.5",
+	"apprise~=1.9.9",
+	"beautifulsoup4~=4.14.3",
+	"celery~=5.6.3",
+	"croniter~=6.2.2",
+	"defusedxml~=0.7.1",
+	"django~=6.0.4",
```
FuzzyGrim, line 14: This is not just a packaging migration. It upgrades Django from 5.2.13 to 6.0.4, while the PR says dependency upgrades are not part of the change.

pyproject.toml
```diff
@@ -1,11 +1,57 @@
+[project]
+name = "yamtrack"
+version = "0.25.0"
```
FuzzyGrim, line 3: Is this needed?

FuzzyGrim (commented):
Thanks for the pull request; I left some comments. Also, could you try running the Docker image created from these changes?

pyproject.toml
daviddavis: It's a required field. It can be set to arbitrary value (`version = "0"`) or I think the better solution would probably be to set `dynamic = ["version"]`, which should allow the field to be omitted.

daviddavis:
I'd love to have this for development. Hope it's ok if I continue this work (https://github.com/FuzzyGrim/Yamtrack/pull/1434).

FuzzyGrim:
Closing in favor of #1434. Thanks @minerop5000 for the initial work!

PR #1316, opened 2026-03-31
Fix Incremental sync with trakt 

### LINKED ISSUE #655: [BUG] Import not importing new watched episodes
First of all, I don't really know if this should be considered a bug of a feature request, but, from my perspective, the sync only new items mode when importing from Trakt can be confusing, so that's my reason to create this issue as a bug. If the current flow is what is expected and designed, can you please explain the reasons behind the decision? 

**Describe the bug**
Right now when importing from Trakt using the sync new items mode, when an in progress show is to be imported and it is already created on the database, the import process skips this show. But there can be new watched episodes from that show, and these new items (watched episodes) are never imported.

**To Reproduce**
Asuming a show not yet tracked on Yamtrack.
1. On Trakt, watch a few episodes from any show.
2. On Yamtrack, import from Trakt using new items mode.
3. On Trakt, wacht more episodes from the previous show.
4. On Yamtrack, import from Trakt using new items mode again.
5. The watched episodes from step 3 are not imported.

**Expected behavior**
The new watched episodes from in progress shows should be added to the database and marked as complete.

**Yamtrack version**:
**Database**: PostgreSQL

darkcoderrises:
Currently incremental sync with trakt, gets all the data from trakt again, which is hugely wasteful. We get only the data from the last time the sync happened. 
Also update the episodes of shows that were watched / rewatched. 
Movies rewatch to also fix.
Fixes #655

### DIFF
--- a/src/integrations/imports/helpers.py
+++ b/src/integrations/imports/helpers.py
@@ -11,12 +11,28 @@ from django.conf import settings
 from django.contrib import messages
 from django.utils import timezone
 from django_celery_beat.models import CrontabSchedule, PeriodicTask
-from simple_history.utils import bulk_create_with_history
+from simple_history.utils import bulk_create_with_history, bulk_update_with_history
 
 import app
-from app.models import MediaTypes
+from app.models import MediaTypes, Sources
 
 logger = logging.getLogger(__name__)
+
+
+def get_last_import_date(user, task_name):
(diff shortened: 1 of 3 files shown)


darkcoderrises:
I am yet to test it, please let me know if the approach is worth it so that I go ahead with testing it.

DdraigX:
Any progress on this?

PR #1319, opened 2026-04-01
feat: add Plex media.rate event support for rating sync

### LINKED ISSUE #1223: [Feature Request] Support "media.rate" event on plex webhook
When I rate media in plex I'd like to see the rating automatically in Yamtrack as well.

trbn1:
Implements support for the Plex media.rate webhook event, allowing users to automatically sync their Plex ratings to Yamtrack.

Changes:
- Added _handle_rating() method to BaseWebhookProcessor for rating updates
- Modified _handle_movie(), _handle_anime() to accept optional rating parameter for playback events
- Added _get_rating_from_payload() helper to extract ratings from payloads
- Added _process_rating() in PlexWebhookProcessor, including anime detection (AniDB, TVDB, TMDB movie, IMDB)
- Expanded MEDIA_TYPE_MAPPING to support Show and Season types for ratings
- Added tests for rating events and playback with ratings

Resolves: #1223

From my limited testing on my own library it seems to work well, but
more testing would be appreciated, in case I missed some edge scenarios.

### DIFF
--- a/src/integrations/tests/test_webhooks_plex.py
+++ b/src/integrations/tests/test_webhooks_plex.py
@@ -675,17 +675,15 @@ class PlexWebhookTests(TestCase):
 
     def test_get_rating(self):
         """Test extraction of rating from payload."""
-        payload_with_rating = {
-            "Metadata": {"userRating": 8}
-        }
+        payload_with_rating = {"Metadata": {"userRating": 8}}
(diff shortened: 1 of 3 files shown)


trbn1:
I went ahead with the refactor to extract common functions in into `base.py`, if that is not something we want under this PR, then I can keep the previous version with duplicated logic.

PR #1327, opened 2026-04-06
fix: use absolute URL for Trakt OAuth token refresh redirect_uri

### LINKED ISSUE #1023: [BUG] oauth with trakt fails
**Describe the bug**
Try to connect to a private trakt profile which requieres authentification

**To Reproduce**
Settings -> Import Data -> Import Trakt -> Import private profile -> Connect with trakt & import

**Expected behavior**
Request to enter credentials

**Screenshots**
OAuth error
The requested redirect uri is malformed or doesn't match client redirect URI.

**Yamtrack version**: Version: v0.24.8
**Database**: SQLite (default) or PostgreSQL

gmanaud:
## Summary
- Fix `get_access_token()` to build the `redirect_uri` using `settings.URLS[0] + reverse()` instead of `settings.BASE_URL` alone
- This ensures the redirect_uri matches the one sent during initial OAuth authorization via `request.build_absolute_uri()`
- Without this fix, deployments on a subpath (`BASE_URL=/yamtrack`) get a 400 from Trakt's token endpoint

Fixes #1023

## Test plan
- Deploy Yamtrack on a subpath (e.g. `/yamtrack`) behind a reverse proxy
- Configure Trakt OAuth and perform an initial import
- Wait for the access token to expire and trigger a refresh (or manually invalidate it)
- Verify the refresh succeeds instead of returning 400

### DIFF
--- a/src/integrations/imports/trakt.py
+++ b/src/integrations/imports/trakt.py
@@ -91,7 +91,7 @@ def get_access_token(encrypted_refresh_token):
         "client_secret": settings.TRAKT_API_SECRET,
         "refresh_token": decrypted_token,
         "grant_type": "refresh_token",
-        "redirect_uri": f"{settings.URLS[0]}{reverse('import_trakt_private')}",
+        "redirect_uri": f"{settings.URLS[0].rstrip('/')}{reverse('import_trakt_private')}",
     }
 
     try:
--- a/src/integrations/imports/trakt.py
+++ b/src/integrations/imports/trakt.py
@@ -91,7 +91,7 @@ def get_access_token(encrypted_refresh_token):
         "client_secret": settings.TRAKT_API_SECRET,
         "refresh_token": decrypted_token,
         "grant_type": "refresh_token",
-        "redirect_uri": f"{settings.BASE_URL}/import/trakt/private",
+        "redirect_uri": f"{settings.URLS[0]}{reverse('import_trakt_private')}",
     }
 
     try:


FuzzyGrim:
How do you know this is the root cause of the linked issue?

gmanaud:
Because I reproduced and fixed it on my own instance:

I maintain a YunoHost package for Yamtrack ([YunoHost-Apps/yamtrack_ynh](https://github.com/YunoHost-Apps/yamtrack_ynh)) which deploys Yamtrack on a subpath (e.g. /yamtrack), so BASE_URL=/yamtrack in the .env. With this configuration:

- The initial authorization succeeds because trakt_oauth() uses request.build_absolute_uri(reverse("import_trakt_private")), producing https://example.com/yamtrack/import/trakt/private.

- The token refresh fails with 400 Bad Request because get_access_token() uses f"{settings.BASE_URL}/import/trakt/private", producing "/yamtrack/import/trakt/private" — a relative path, missing the scheme and host, so the redirect_uri sent with a refresh_token grant doesn't match with the one sent at authorization, and the server rejects the request.

I patched get_access_token() on my production server to match the authorization URL, restarted the services, and the refresh now succeeds.

I linked #1023 because the symptoms match (400 on token endpoint, redirect_uri mismatch), but #1023 describes several related problems — e.g. HTTP vs HTTPS when behind a reverse proxy, missing state parameter, dropped port numbers — which look like different root causes (reverse-proxy headers, etc.) that this PR does not address. If you prefer, I can drop the Fixes #1023 line and just reference it ("Related to #1023"), since this PR only fixes one of the code paths described there.

FuzzyGrim:
Closing in favor of https://github.com/FuzzyGrim/Yamtrack/commit/3d01a88503cb070eb6ce1c832c3dbde67b5ba474

Issue #1198, opened 2026-02-14
[Feature Request] Allow to hide upcoming episodes from "In Progress"


foux:
**Describe the solution you'd like**  
I was trying to replicate Trakt Up Next feature. Sorting "In Progress" by recent wil almost do it, but with an issue : it will keep upcoming episodes in the list.

It would be nice to be able to hide those, so that we only see what we have to see next on the list.

cdemi:
I think this is the only feature I'm missing from Trakt

neilsb:
This issue should be closed now as it was resolved in #1371

sboddy:
I've seen one wrinkle that it might be nice to fix first.
I have a new season set to planning. I haven't yet started watching it, but Episode 1 has been released, and Episode 2 is in the future. To my mind this season should still be displayed in Planning when `Hide unreleased` is active.

Once a whole season has been completely released, it appears in the Planning section.

sboddy:
#1702

PR #1371, opened 2026-04-22
Add option to hide 'Upcoming" episodes from the home page 

thespbgamer:
Still needs a bit of refinement (namely on the icons), besides that, I think the video is self explanatory 

https://github.com/user-attachments/assets/0dfa3087-07f6-4d98-bc36-3763fd55b192


src/app/views.py
```diff
@@ -36,49 +36,180 @@
...
+                "total": len(filtered_items),
+            }
+    return filtered_media_types
+
+
+def _paginate_home_media_types(media_types, items_limit, page_start=0):
+    """Paginate already-grouped home media entries."""
+    paginated_media_types = {}
+    for media_type, media_list in media_types.items():
+        items = media_list["items"][page_start : page_start + items_limit]
```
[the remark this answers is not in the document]
thespbgamer: None issue

FuzzyGrim:
What's the difference between Incoming and Planning?

thespbgamer:
> What's the difference between Incoming and Planning?


Well:
- Let's say you're watching show X, you're watching it weekly (when it releases) and you're caught up 
- Let's say you're watching show Y (it's an old show, everything is available), and your half way there
- Let's say you're planning to watch show Z (but you don't know when)

With this separation you have a clear distinction between:
- Things you are watching and **can't** watch (in progress but upcoming) = show X
- Things you are watching and can watch (in progress) = show Y
- Things you're planning to watch but don't know when (planning) = show Z

thespbgamer:
Also this will semi close https://github.com/FuzzyGrim/Yamtrack/issues/1198

FuzzyGrim:
Okay, I understand the feature request of not wanting to see upcoming episodes listed. But I'm not sure about showing them in a section below. It feels very cluttered, like if you track multiple types of media, the page would be very long to scroll.

thespbgamer:
You're not wrong, was trying to make it similar to what I used (aka Trakt)

thespbgamer:
Do you have any suggestions on what should be done, I know this issue is a thing at least for me and a couple of friends, but not sure what the best alternative is, of the top of my head, maybe a 'collapse' option for the sections so we don't need to scroll, wdyt @FuzzyGrim ?

(sorry if you don't want/like the tag)

amadejpapez:
My thoughts are that this should be something simpler. I remember Trakt at least in lists had a filter that could filter out Unreleased items which I used extensively when browsing my watchlist to find stuff available now.

I would suggest something similar here. Somewhere at the top (alongside of the sorting options) or under the In Progress title - a filter to hide Unreleased items. Which could work across all of the components on that view - hiding not yet released movies, hiding shows with all released episodes watched, hiding unreleased games and so on in their respective existing lists.

So I think this should be handled as a filter not in settings and with separate components list as is currently done.

Will wait and see what Fuzzy thinks. But it may be address his thoughts about a separate list component.

thespbgamer:
> My thoughts are that this should be something simpler. I remember Trakt at least in lists had a filter that could filter out Unreleased items which I used extensively when browsing my watchlist to find stuff available now.
> 
> I would suggest something similar here. Somewhere at the top (alongside of the sorting options) or under the In Progress title - a filter to hide Unreleased items. Which could work across all of the components on that view - hiding not yet released movies, hiding shows with all released episodes watched, hiding unreleased games and so on in their respective existing lists.
> 
> So I think this should be handled as a filter not in settings and with separate components list as is currently done.
> 
> Will wait and see what Fuzzy thinks. But it may be address his thoughts about a separate list component.

did these changes, forgot to mention, video update in PR description

src/app/tests/views/test_home.py
```diff
@@ -202,6 +218,63 @@ def test_home_view_with_sort(self):
...
+        """Test home view can hide unreleased media in existing sections."""
+        response = self.client.get(reverse("home") + "?hide_unreleased=1")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(response.context["hide_unreleased"])
+
+        sections_by_key = {
+            section["key"]: section for section in response.context["home_sections"]
+        }
+        self.assertNotIn("incoming", sections_by_key)
```
FuzzyGrim: Incoming section was removed with the update

src/app/views.py
```diff
@@ -114,8 +244,20 @@ def progress_edit(request, media_type, instance_id):
         media.refresh_from_db()
         prefetch_related_objects([media], "episodes")
 
+    if hide_unreleased and home_status == Status.IN_PROGRESS.value:
+        BasicMedia.objects.annotate_max_progress([media], media_type)
```
FuzzyGrim: Is this needed? The `annotate_max_progress()` from `get_media_prefetch` invoked by `helpers.get_owned_media_or_404` should work.

src/app/views.py
```diff
@@ -114,8 +244,20 @@ def progress_edit(request, media_type, instance_id):
...
+        BasicMedia.objects._annotate_next_event([media])
```
FuzzyGrim, line 270: `media.refresh_from_db()` from L244 above clears the cached item relation, so _annotate_next_event's getattr(media.item, 'prefetched_events', []) always falls back to [] for Season items, meaning the auto-hide-card behavior never triggers for TV/Season progress edits. The added test only covers Anime; can you add a Season-specific test and make the events prefetch survive?

src/app/tests/views/test_home.py
```diff
@@ -202,6 +218,63 @@ def test_home_view_with_sort(self):
...
+
+        in_progress_section = sections_by_key[Status.IN_PROGRESS.value]
+        planning_section = sections_by_key[Status.PLANNING.value]
+
+        self.assertEqual(in_progress_section["count"], 1)
+        self.assertIn(MediaTypes.ANIME.value, in_progress_section["media_types"])
+        self.assertNotIn(MediaTypes.SEASON.value, in_progress_section["media_types"])
+        self.assertEqual(planning_section["count"], 1)
+
+    def test_home_view_htmx_load_more_preserves_hide_unreleased(self):
```
FuzzyGrim, line 273: This test fails

src/app/views.py
```diff
@@ -44,52 +44,180 @@
...
+
+
 @require_GET
 def home(request):
     """Home page with media items in progress and planning."""
     sort_by = request.user.update_preference("home_sort", request.GET.get("sort"))
     media_type_to_load = request.GET.get("load_media_type")
-    status_to_load = request.GET.get("load_status", Status.IN_PROGRESS.value)
+    section_to_load = request.GET.get("load_status", Status.IN_PROGRESS.value)
+    hide_unreleased = request.GET.get("hide_unreleased") == "1"
```
FuzzyGrim: We should save this as a user preference in settings.

FuzzyGrim (commented):
Thanks for the video, looks great!

src/app/tests/views/test_home.py
thespbgamer: True, fixed.

src/app/views.py
thespbgamer: Not needed, UX improved due to this.

src/app/views.py
thespbgamer: Tests added to cover this.

src/app/tests/views/test_home.py
thespbgamer: Fixed, 100% my bad, did things 'too fast' before, just force commited things.

src/app/views.py
thespbgamer: Re-added setting save for this and migration

amadejpapez:
I have started using this and it is really nice! What I have noticed is though that this should probably get added to other tabs like Movies/TV Shows/... and so on as well not just the home page.

Nicely removes unreleased stuff from being in my face as i cannot do nothing about them in the home page. But if i go into Movies tab and want to see what i want to watch next from there in a more full view, I cannot filter out unreleased stuff there.

PR #1373, opened 2026-04-23
feat: rootless container & small improvements

### LINKED ISSUE #1120: Running Yamtrack as non-root and read-only with docker
Is it possible to run the container this way? I tried the following compose.yaml:

```yaml
include:
  - ../compose.common.yaml
  - ../traefik/compose.yaml
  - ../valkey/compose.yaml

services:
  yamtrack:
    container_name: yamtrack
    image: ghcr.io/fuzzygrim/yamtrack:0.24.11

    user: $PUID:$PGID
    read_only: true
    security_opt:
      - no-new-privileges
    cap_drop:
      - ALL

    networks:
      - internet  # For metadata.
      - traefik
      - valkey

    environment:
      SECRET: $YAMTRACK_SECRET
      REDIS_URL: redis://valkey:6379
      TZ: $TZ

    expose:
      - &port 8000

    labels:
      traefik.enable: true
      traefik.http.routers.yamtrack.rule: Host(`yamtrack.$DOMAIN`)
      traefik.http.services.yamtrack.loadbalancer.server.port: *port 

      glance.name: &name Yamtrack
      glance.icon: &icon sh:yamtrack
      glance.url: https://yamtrack.$DOMAIN
      glance.category: frontend/users
      glance.description: Media tracker

      wud.watch: true
      wud.display.name: *name 
      wud.display.icon: *icon 
      wud.tag.include: ^\d+\.\d+\.\d+$$
      wud.link.template: https://github.com/fuzzygrim/yamtrack/releases/tag/v$${major}.$${minor}.$${patch}

    tmpfs:
      - /tmp:noexec

    volumes:
      # - ./cache:/app/cache:rw
      - ./state:/yamtrack/db:rw

    restart: unless-stopped
```

But I get the following error.
```
yamtrack  | Operations to perform:
yamtrack  |   Apply all migrations: account, admin, app, auth, contenttypes, django_celery_beat, django_celery_results, events, lists, sessions, socialaccount, users
yamtrack  | Running migrations:
yamtrack  |   No migrations to apply.
yamtrack exited with code 10 (restarting)
```

_Originally posted by @UtkarshVerma in https://github.com/FuzzyGrim/Yamtrack/discussions/1070_

MrRubberDucky:
This PR makes container run as a rootless user by default. Fixes #1120 and replaces PR #1128 

Following files were modified:
- supervisord.conf
- nginx.conf
- Dockerfile
- entrypoint.sh
- /src/config/settings.py

## Changes

Container now runs as user & group `abc:abc` with UID & GID of `1000`, similar to aforementioned PR. It also takes care of permission inside the container, ensuring that abc user can write to directories it needs. All logs get written to `/tmp` as rootless user can't access `/dev/stdout`, `/dev/stderr` - supervisor writes stdout and stderr of apps to `/tmp/appname-stdout---supervisord-randomstring.log` so I suppose it's not that big of an issue.

There's one problem and it's with nginx trying to write an `/var/lib/nginx/logs/error.log` all the time but it just seems to be a hard-coded value and any errors are still logged to supervisord stdout log files if they occur. Since this isn't considered fatal, I see it as a pointless thing to try to fix.

Added `USE_X_FORWARDED_HOST` to `/src/config/settings.py` as otherwise I couldn't get django to trust Caddy's `X Forwarded For` header. This will be more useful for people that will reverse proxy their Yamtrack container to the web and want SSO to work properly. Without it, SSO redirect gets built wrongly and instead of the domain from the headers, it used LAN IP address for redirect, causing invalid callback URL error.

Modified `nginx.conf` to listen on both IPv4 and IPv6 by default. Currently the Dockerfile creates another IPv6 configuration file then uses an undocumented environment variable to switch between them, which seems pointless considering that nginx supports dual-stack connections just fine. Additionally, all relevant temp and log files are written to `/tmp` to allow it to run rootless without barrage of permission errors or having to constantly mount everything to be owned under current user.

entrypoint.sh was reduced to just two simple commands: the simple migration python command and supervisord exec.

`PGID` and `PUID` are unused variables now.

## Potential problems arising from this PR

SQLite installs with bind volumes will need to `chown` their volumes to be owned by rootless user inside container, otherwise they will get either `out of memory` (happens when SQLite can't write to the directory, or to the database), or errors such as `access denied`. User ownership doesn't matter for any other directory as the permissions are allowing enough with them being `r-x`.

## Did you test it?

Yup. It works fine, nothing else to say really.

## What's the benefit?

You can now run the container as any rootless user and as read only, as long as you fix relevant directory ownership yourself. This gets rid of the convenience aspect but in turn it increases security as container init and processes running under it are ran via non-root user with strict permissions that prevent modification to `entrypoint.sh` and anything copied from `/src/*`

If you will be using external database, you can run it as read only easily without suffering through permission errors. Here's my current Quadlet as an example that runs as my user with randomized user namespace and read only.

```bash
[Unit]
Description=Yamtrack - self hosted media tracker for movies, tv shows, anime, manga, video games, books, comics, and board games.
Requires=postgres-yamtrack.service
Requires=valkey-yamtrack.service
After=postgres-yamtrack.service
After=valkey-yamtrack.service

[Install]
WantedBy=default.target

[Service]
Restart=on-failure
SystemCallArchitectures=native
MemoryDenyWriteExecute=false

[Container]
Image=localhost/yamtrack:test
ContainerName=yamtrack
EnvironmentFile=%h/Environments/Yamtrack/.env
AutoUpdate=registry
DropCapability=ALL
NoNewPrivileges=true
ReadOnly=true
Network=Yamtrack.network
# Internal networks
Network=Postgres-Yamtrack.network
Network=Valkey-Yamtrack.network
PublishPort=9103:8000
User=%U:%U
UserNS=auto:size=1002
```

...and no, this wasn't written by LLM before anyone asks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,11 +9,9 @@ ARG VERSION=dev
 ENV VERSION=$VERSION
 
 COPY ./requirements.txt /requirements.txt
-COPY ./entrypoint.sh /entrypoint.sh
 COPY ./supervisord.conf /etc/supervisord.conf
+COPY --chmod=555 ./entrypoint.sh /entrypoint.sh
 COPY ./nginx.conf /etc/nginx/nginx.conf
-# Generate a copy of the nginx config with IPv6 support.
-RUN sed 's/listen 8000;/listen 8000; listen [::]:8000;/' /etc/nginx/nginx.conf > /etc/nginx/nginx.ipv6.conf
 
 WORKDIR /yamtrack
 
@@ -22,15 +20,13 @@ RUN apk add --no-cache nginx shadow \
     && pip install --no-cache-dir supervisor==4.3.0 \
     && rm -rf /root/.cache /tmp/* \
     && find /usr/local -type d -name __pycache__ -exec rm -rf {} + \
-    && chmod +x /entrypoint.sh \
-    # create user abc for later PUID/PGID mapping
-    && useradd -U -M -s /bin/sh abc \
-    # Create required nginx directories and set permissions
-    && mkdir -p /var/log/nginx \
-    && mkdir -p /var/lib/nginx/body
+    && useradd -U -M -u 1000 -s /bin/sh abc \
+    && mkdir -p /yamtrack/db /yamtrack/staticfiles /var/cache/nginx \
+    && chown -R abc:abc /yamtrack/db /yamtrack/staticfiles /var/cache/nginx
 
-# Django app
-COPY src ./
+COPY --chmod=555 --chown=abc:abc src ./
+
+USER abc:abc
 RUN python manage.py collectstatic --noinput
 
 EXPOSE 8000
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -4,16 +4,4 @@ set -e
 
 python manage.py migrate --noinput
 
-PUID=${PUID:-1000}
-PGID=${PGID:-1000}
-
-groupmod -o -g "$PGID" abc
-usermod -o -u "$PUID" abc
-
-chown abc:abc /yamtrack
-chown -R abc:abc db
-chown -R abc:abc staticfiles
-chown -R abc:abc /var/log/nginx
-chown -R abc:abc /var/lib/nginx
-
 exec /usr/local/bin/supervisord -c /etc/supervisord.conf
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,17 +1,25 @@
-user abc;
-worker_processes auto;
+worker_processes  auto;
+error_log /tmp/error.log;
 pid /tmp/nginx.pid;
-error_log /dev/stderr warn;
 
 events {
-    worker_connections 1024;
+    worker_connections  1024;
 }
 
 http {
+    proxy_temp_path /tmp/proxy_temp;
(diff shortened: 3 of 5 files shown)


MrRubberDucky:
Ah the clanker had a field trip here, we truly live in the future. Not gonna resolve these (as I don't really know how and I don't wanna accidentally nuke this PR with LLM changing stuff 'cus it misunderstood what I've said) as they're non-issues and are done the way they are due to supervisord logic doing funky things while trying to access `/dev/stdout` / `/dev/stderr` that makes it error out with `EACCES` on rootless user, I may know that's the case because I've tried it 😆 

IPv6 is generally fine anyways as rootless networks do both IPv4 and IPv6 even if it's disabled on host and any IPv4 connection is properly resolved because nginx & rootful / rootless networking is just clever like that. It could be reverted though at the request of a maintainer, or changed back to how it was, I don't mind.

This would also replace #1264 as I set entrypoint.sh permissions here to `555`.

Edit: I've been thinking if it's worthwhile dumping logs to /tmp/logs instead, or finding out how nginx rootless image does it if logs being displayed on `docker logs` is an important feature to have. I'll look into it in case it's wanted.

PR #1378, opened 2026-04-25
feat(api): Implement dfr-spectacular for OpenAPI spec generation

66Bunz:
This PR adds drf-spectacular (https://github.com/FuzzyGrim/Yamtrack/issues/488#issuecomment-4072954234) to generate and validate an OpenApi spec for the API.

This is an initial implementation that will still need work after this PR is merged, because until the `api` and `app` implementations are not merged, a lot of elements needs to be duplicated and hard-coded (see below). 

## What's new

-  Add new `uv run manage.py generate_openapi --validate` management command to generate and validate API spec in the pre-commit hooks
- Add schemas to serializers
- Add routes `api/v1/schema` and `api/v1/docs` for schema and Swagger documentation views

## Next steps

The next steps are the following:

- Start adding API documentation in the `docs/` folder
- **Start merging `app` and `api` implementations to have only one source of truth**
- Completely move webui implementations to `api` logic


src/config/settings.py
```diff
@@ -352,6 +351,31 @@ def secret(key, default=undefined, **kwargs):
...
+SPECTACULAR_ENABLE_SERVE = config(
+    "SPECTACULAR_ENABLE_SERVE",
+    default=DEBUG,
+    cast=bool,
+)
+
+SPECTACULAR_SETTINGS = {
+    "TITLE": "Yamtrack API",
+    "DESCRIPTION": "OpenAPI schema for Yamtrack's API",
+    "VERSION": "0.0.25",
```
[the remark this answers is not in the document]
66Bunz: Maybe when the API will be stable, not now

src/api/views.py
```diff
@@ -719,9 +1914,152 @@ def get(self, request, list_id, item_id):
...
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                required=False,
+                description=(
+                    "Comma-separated list of media types to exclude (excluded"
+                    "by default are seasons and episodes)."
+                ),
+                enum=[media_type.value for media_type in MediaTypes],
+            ),
+            MediaTypeCompleteParam,
```
FuzzyGrim: `MediaTypeCompleteParam` seems to be required here, don't think it makes sense

src/api/schemas.py
```diff
@@ -0,0 +1,148 @@
...
+    required=True,
+)
+
+SourceParam = OpenApiParameter(
+    name="source",
+    type=OpenApiTypes.STR,
+    location=OpenApiParameter.PATH,
+    description="Source of media item data for import operations.",
+    required=True,
+    enum=SOURCES_VALID_LIST,
```
FuzzyGrim, line 195: This excludes manual items.

src/api/helpers.py
```diff
@@ -162,6 +165,19 @@
...
+    Sources.MAL.value,
+    Sources.MANGAUPDATES.value,
+    Sources.IGDB.value,
+    Sources.OPENLIBRARY.value,
+    Sources.HARDCOVER.value,
+    Sources.COMICVINE.value,
+    Sources.BGG.value,
+]
+
+SOURCES_COMPLETE_VALID_LIST = [*SOURCES_VALID_LIST, "manual"]
```
FuzzyGrim, line 179: This hard codes `manual`.

README.md
```diff
@@ -79,6 +79,22 @@ The default Compose file uses SQLite, which is enough for most personal installs
...
 
+### OpenAPI schema generation
+
+The API spec in [openapi.yaml](openapi.yaml) can be generated automatically from Django/DRF using [`drf-spectacular`](https://github.com/tfranzel/drf-spectacular).
+
+From the project root, run:
+
+```bash
+cd src
+python manage.py generate_openapi --validate
```
FuzzyGrim: Should be `uv run...`, but maybe this section should be at `docs/development/`

src/api/helpers.py
FuzzyGrim, line 1: Can we reduce these helper constants to one source of truth? `MEDIA_STATUS_CHOICES` can likely come from the existing status choices, `SOURCES_*_VALID_LIST` should be derived from Source choices...

Models:

```
django.apps.apps.get_models()
django.apps.apps.get_model()
```

src/api/serializers.py
```diff
@@ -573,20 +591,665 @@ def to_representation(self, instance):  # noqa: ARG002
...
+    item = ItemSerializer()
+    item_id = ItemIdField(source="item", allow_null=True, required=False)
+    parent_id = ParentIdField(source="item", allow_null=True, required=False)
+    tracked = serializers.BooleanField()
+    created_at = serializers.DateTimeField(allow_null=True, required=False)
+    score = serializers.FloatField(allow_null=True)
+    status = MediaStatusChoiceField()
+    progress = serializers.IntegerField(allow_null=True)
+    progressed_at = serializers.DateTimeField(allow_null=True)
+    start_date = serializers.DateField(allow_null=True)
```
FuzzyGrim: Start and end dates are actually `DateTimeField`

src/api/serializers.py
```diff
@@ -112,243 +193,124 @@ def __init__(self, status_value):
...
+    item = ItemSerializer()
+    item_id = ItemIdField(source="item")
+    parent_id = ParentIdField(source="item")
+    tracked = serializers.BooleanField()
+    created_at = serializers.DateTimeField(allow_null=True, required=False)
+    score = serializers.FloatField(allow_null=True)
+    status = MediaStatusChoiceField()
+    progress = serializers.IntegerField(allow_null=True)
+    progressed_at = serializers.DateTimeField(allow_null=True)
+    start_date = serializers.DateField(allow_null=True)
```
FuzzyGrim: Episodes don't have a start date nor notes. The end date should be a `DateTimeField`

FuzzyGrim (changes requested):
Thanks for adding generated OpenAPI support. I like the direction, but I’m worried this PR replaces the hand-written `openapi.yaml` with a lot of manually maintained schema code spread across `src/api/serializers.py` and `src/api/views.py`.

A lot of the serializers now look like schema-only mirrors: they declare fields for drf-spectacular, but then override `to_representation()` and manually return dictionaries. That gives us two sources of truth: the declared serializer fields and the actual runtime response. If they disagree, the generated OpenAPI becomes wrong. For example, some serializers declare fields as required/non-null while the returned dict can emit `None`.

Same concern for request serializers: if we define request serializers for docs, but views still parse `request.data` manually, the docs can drift from real validation. Where possible, the request serializer should actually be used:

```python
serializer = ListCreateRequestSerializer(data=request.data)
serializer.is_valid(raise_exception=True)
data = serializer.validated_data
```

I think we can make this much more minimal:

- Use `ModelSerializer` for model-backed responses and let DRF/drf-spectacular infer fields from `app/models.py`.
- Only override fields that differ from the model/API contract.
- Use explicit `Serializer` classes for non-model payloads like pagination, info responses, search results, and request bodies.
- Use `DictField`, `JSONField`, or `OpenApiTypes.OBJECT` for flexible provider metadata instead of modeling every provider-specific `details` shape upfront.
- Keep `extend_schema` only for cases where inference is wrong or incomplete.
- Centralize repeated auth/error/pagination responses instead of copying large `400`/`403`/`404`/`500` examples into every endpoint.

The goal should be for the generated schema to come from the same objects that validate/serialize the API at runtime.

66Bunz:
Everything you said is true, but since the api still heavily relies on the old methods of getting and putting data in the db, this was a PR to just implement the drf-spectacular framework, even if everything is still fragmented between api documentation and new and old implementations.
The plan in the next PR is to move a lot of the implementations to the API side, and make everything use the api only.
I completely forgot to create the PR body to tell what the next steps were, I'm sorry.
I'll do the changes

src/api/serializers.py
66Bunz: I know, but since all the medias should return an uniform response, the start date is set to the created time, and the notes are set to None.

src/api/helpers.py
66Bunz: It'll be until I merge the `app` in the `api`. Right now pretty much everything is still hardcoded, in the next pr I'll start merging stuff and everything will be more tidy.

src/api/helpers.py
66Bunz: same as https://github.com/FuzzyGrim/Yamtrack/pull/1378#discussion_r3441250862

66Bunz:
> Centralize repeated auth/error/pagination responses instead of copying large `400`/`403`/`404`/`500` examples into every endpoint.

@FuzzyGrim since in every view's method an error response can have different examples, do you mean something like this:

```python
500: OpenApiResponse(
    ApiErrorResponseSerializer,
    description="Internal server error",
    examples=[Example500Event],
),
```
where only the example is set in the `schemas.py` file

or 

```python
500: internal_error_response_event,
```
where the whole event response for error `500` is set, so we'll have various `internal_error_response_*` 

or

```python
500: internal_error_response
```
if we don't care about examples for the responses and just have a generic response?

---

Another thing that can be done is creating a builder function in `schemas.py` that combines the examples with the response, something like this

```python
500: get_response(500, "event_delete")
```

```python
def get_response(error, route):
    return OpenApiResponse(
        ApiErrorResponseSerializer,
        description=descriptions[error]
        examples=examples[route]
    )
```

This is obviously where the response is different every time.

FuzzyGrim:
For 401/403/404/500 I would go with generic responses

PR #1390, opened 2026-05-02
Add support for X-Forwarded-* headers for better reverse proxy support.

didli789:
Solves #1023 in a better way than #1327 that also works in other imports, without the security problem of spoofing possibility, by having the user only enable it when Yamtrack is used behind another reverse proxy.


didli789:
Added USE_X_FORWARDED_PROTO and USE_X_FORWARDED_PORT, and also USE_X_FORWARDED that sets the default for the other three.

Issue #1407, opened 2026-05-10
[Feature Request] Support for concerts


CallMeAlexO:
**Describe the solution you'd like**  
I'd like to track concerts. No other tracker supports this, and it's a damn shame. Some mainstream paid/closed-source trackers do offer this functionality.

**Describe alternatives you've considered**  
Lately, my wife is using [repov](https://play.google.com/store/apps/details?id=com.kilo.repov&hl=en-US). But they've added an ungodly amount of spam and social features. Only real difference between Yamtrack and repov for us is the concert tracking. It'd be nice to move some of that inside the house.

**Additional context**  
setlist.fm is a free API that everyone has access to. It should be possible to just add support for that.

PR #1408, opened 2026-05-10
feat: Add support for concerts

CallMeAlexO:
# What is this?

Baseline solution for #1407. Very barebones, let me know if there's anything that could be improved :)

To actually fetch setlists, you'll need an API key - `SETLISTFM_API_KEY`. I don't know who owns the default sample API key, so I didn't add one. But if this is a requirement for the PR, let's add it.

To add a concert, simply search for the artist and scroll past the results. Unfortunately Yamtrack does not yet support date-based searches, filtering, and ordering in search results -  which can be kind of hard for concerts. But it's functional.

Also, setlist.fm does not have artwork associated with setlists so you'll have to add your own. I've added the ability to add a cover using a URL.

# What I didn't do (and why)

`setlist.fm` has a very nice search feature. Where you can put something like `"london grammar paris"` and it will detect that you're searching for the artist "London Grammar" for setlists in Paris. However, this search endpoint is not exposed in the official API. It's possible to implement, but there are so many competing strategies to do so, which I think goes beyond the scope of just getting concerts to show up on the tracker. So I decided against it.

Another thing that could be nice is grouping setlists of the same date and same venue together (think festivals, opening acts, etc). However, this creates a few issues with how Yamtrack works now. I'd have to remove the artist from the details and add artists in a manner similar to how actors in Movies are. Not to mention, the artist's image is not available on `setlist.fm`, so we'd need a separate API call for those. For now, I decided to just get it barebones. Perhaps this could be done in the future.  

---

# Screenshots

Search results:
search results

Details view:
details view in search

Details in tracker:
image


66Bunz:
For concerts with no images, what about displaying the artists logo/image?

CallMeAlexO:
> For concerts with no images, what about displaying the artists logo/image?

That would require a separate API call for artists, as setlist.fm does not have artist images.

PR #1430, opened 2026-05-16
Fall back to TV show image when season image is missing

daviddavis:
When TMDB does not return a poster path for a season, the season image is set to the placeholder `IMG_NONE`. This change makes `enrich_season_with_tv_data` substitute the parent TV show's image in that case, so seasons that lack their own art render the show's poster instead of the placeholder.

Since enrichment runs whenever season metadata is fetched/cached via `tv_with_seasons`, all downstream consumers (views, models, calendar, webhooks, imports) inherit the fallback automatically.

Added unit tests covering both the fallback case and the case where the season already has its own image.


daviddavis:
I think the issue I am mostly facing I have fixed in https://github.com/FuzzyGrim/Yamtrack/pull/1432

PR #1432, opened 2026-05-16
Refresh Item.image on details page when missing

### LINKED ISSUE #1365: [BUG] TV shows in progress missing thumbnails
**Describe the bug**
Some TV shows don’t show thumbnails in the In Progress section. Movies and anime display thumbnails correctly, but certain TV shows are missing them.

**To Reproduce**
Try adding any of these shows in the screenshot.

**Expected behavior**
Some shows have thumbnails but some don't.

**Screenshots**

Image

Image

**Yamtrack version**: v0.25.2
**Database**: SQLite

daviddavis:
## Summary

The home page renders `media.item.image` directly from the database, which is only written when the `Item` is first created. The details views (`media_details`, `season_details`) fetch metadata live from the provider, so they can show a different — usually correct — image. This is most visible for TV seasons where the poster wasn't available at creation time and never gets backfilled, leaving home-page tiles without a thumbnail.

This PR adds a tiny helper that opportunistically refreshes `Item.image` whenever a details view already has freshly-fetched metadata and the stored image is empty or `IMG_NONE`. No extra API calls — it reuses the metadata the view fetches anyway, and performs a one-column `UPDATE` only when needed.

Applied at three call sites:
- `media_details` — refreshes the current tracked item's image.
- `season_details` — same, for seasons.
- `enrich_items_with_user_data` — also covers tracked items surfaced via related sections (e.g. the seasons listed on a TV details page, recommendations, collections).

## Scope / non-goals
- Only refreshes when the stored image is missing (`""` or `IMG_NONE`). It will not overwrite a different-but-valid stored image, so legitimate edits/imports from other sources are preserved.
- It will not heal items the user never opens; a periodic sync task (e.g. via TMDB's `/changes` endpoint as suggested in the issue thread) would be the complementary fix for that.

## Tests
Added two cases in `test_media_details.py` covering:
- Item with `IMG_NONE` gets updated after visiting the details page.
- Item with an existing image is left alone.

Fixes #1365.

### DIFF
--- a/src/app/helpers.py
+++ b/src/app/helpers.py
@@ -12,7 +12,7 @@ from django.utils import timezone
 from django.utils.encoding import iri_to_uri
 from django.utils.http import url_has_allowed_host_and_scheme
 
-from app.models import BasicMedia, MediaTypes, Status
+from app.models import BasicMedia, Item, MediaTypes, Status
 
 YEAR_ONLY_PARTS = 1
 YEAR_MONTH_PARTS = 2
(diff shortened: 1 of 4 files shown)


FuzzyGrim:
Thanks for the PR! This is something that's been on my to-do list for a long time.

PR #1466, opened 2026-05-27
fix: episode watch modal prefill dates using local timezone

### LINKED ISSUE #1374: [BUG] Episode tracking modal uses wrong time
**Describe the bug**
When you go to "Track Episode" on a show the pre-filled time used follows the timezone set on the server and not the local time of the user. Additionally the time used is cached on page load as opposed to on modal opening which is what happens for movies. Would potentially be nice for the time to not be cached at all for both movies and episodes but that's more a feature request.

**To Reproduce**
For timezone: Set timezone of the server to something other than your own. Click "Track Episode" for any episode in a show.

For caching issue: Open up the season page of any show and wait any amount of time longer than a minute. Click "Track Episode" and the time will pre-fill with the time the page was initially loaded.

**Expected behavior**
Like is done with movies I expect episodes to use local time for it's initial pre-filled value and not the arbitrary time set by the server owner. Although changing the timezone is an option it doesn't solve the problem for other users in other timezones.

For the date caching the behaviour should at least match that of movies. Leaving the page of a season open while binging through it isn't an uncommon occurance. Though better yet might be to not cache the date at all or I suspect the time might drift behind anyways if the modals are currently set to share the cached value (I don't actually know how they are currently designed).

**Yamtrack version**: 0.25.2
**Database**: SQLite

amadejpapez:
Episode watch modal used UTC and not local timezone like the modal for movies. I have also removed `now_plus_minutes()` as I assume that uses server timezone not device local one.

This carries the similar logic as done in `mediaStatusDateHandler.js getCurrentDateTime()`, which does use local timezone.

Fixes https://github.com/FuzzyGrim/Yamtrack/issues/1374

Also setting TIMEZONE via env variables does fix it but it worked inconsistently as movie input read timezone from the client while episode input did not. Please provide feedback if this should work differently.

<br>

**Before:**
Screenshot 2026-05-28 at 12 29 59 AM
Screenshot 2026-05-28 at 12 30 24 AM

<br>

**After:**
Screenshot 2026-05-28 at 12 30 44 AM

### DIFF
--- a/src/app/templatetags/app_tags.py
+++ b/src/app/templatetags/app_tags.py
@@ -1,4 +1,3 @@
-from datetime import timedelta
 from pathlib import Path
 
 from django import template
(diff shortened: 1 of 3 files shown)


FuzzyGrim:
Thanks for your contribution!

PR #1468, opened 2026-05-28
fix: allow generic username to restore PWA shortcuts

### LINKED ISSUE #1460: [BUG] #1251 broke PWA shortcuts
**Describe the bug**
Since #1251 path are preceded by the username, even when connected. This is incompatible with PWA shortcuts since they are supposed to be generic (and currently use the "old" paths).

It ends up in 404 errors.

**To Reproduce**
Update to 0.25.3

**Expected behavior**
Shortcuts should work.
I guess a solution would be to have a fallback to the current logged user if it is not found in the path and/or having a default prefix for the current user ( `/me/` for instance)

**Yamtrack version**:
0.25.3

pogman-code:
To fix #1460, I propose to allow access to the current user media lists using a wildcard username in URLs (`/~/*`).

### DIFF
--- a/src/app/helpers.py
+++ b/src/app/helpers.py
@@ -8,13 +8,14 @@ from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Q
 from django.http import Http404, HttpResponse, HttpResponseRedirect
-from django.shortcuts import redirect
+from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.encoding import iri_to_uri
 from django.utils.http import url_has_allowed_host_and_scheme
 
 from app.models import BasicMedia, Item, MediaTypes, Status
+from users.models import User
 
 YEAR_ONLY_PARTS = 1
 YEAR_MONTH_PARTS = 2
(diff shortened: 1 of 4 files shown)


src/app/helpers.py
```diff
@@ -37,6 +38,14 @@ def get_owned_media_or_404(request, media_type, instance_id, *, prefetch=False):
         raise Http404(msg) from exc
 
 
+def get_user_or_404(request, username):
```
FuzzyGrim, line 43: add a test covering this

src/app/helpers.py
```diff
@@ -37,6 +38,14 @@ def get_owned_media_or_404(request, media_type, instance_id, *, prefetch=False):
...
+    """Return user or raise 404."""
+    if request.user.is_authenticated and username == "~":
+        return request.user
+
+    return get_object_or_404(User, username=username)
```
FuzzyGrim, line 54: Unauthenticated /~/tv returns 404 instead of redirecting to login.

src/app/helpers.py
```diff
@@ -37,6 +38,14 @@ def get_owned_media_or_404(request, media_type, instance_id, *, prefetch=False):
         raise Http404(msg) from exc
 
 
+def get_user_or_404(request, username):
+    """Return user or raise 404."""
```
FuzzyGrim: Docstring omits the special case of "~"

pogman-code:
@FuzzyGrim, all the changes are done, let me know if you want me to change anything else :)

PR #1480, opened 2026-05-31
add option to fill in no date in fill_track_episode

Quitlox:
This PR adds an explicit option to clear the date field in the fill_track_episode dialog.

Unfortunately, not all platforms support clearing the date input field. For example, on my pixel 8 using firefox, I can clear the input field, but it does not allow me to set it ("set" button does not work in picture below). I propose we add an explicit option so that the user can easily clear the field, to add an episode as watched without a date.

WhatsApp Image 2026-05-31 at 21 36 31

I have verified the change on my desktop (linux, firefox) and on my google pixel 8 (firefox). I have ran the test suite just to be sure, all tests pass.


FuzzyGrim:
Thanks for the PR!

PR #1586, opened 2026-07-19
fix: X button on movie track modal not working

amadejpapez:
## Steps

Go to the movie page.
The movie should already have been tracked. Then try to add another entry via `Add new entry`.
A modal opens.
The X button does not work. Click outside of the modal does.

## Where

I was on latest dev branch build. The demo site also has this issue. Other cases seem to work. Except games and books and possible more as such which their `Add new entry` modal also does not close but this should fix those too.

## Fix

The modal seemed to have a mismatch between `trackOpen` and `createOpen` and missing one of them.


amadejpapez:
@FuzzyGrim a small fix that also happens on demo page : )

PR #1608, opened 2026-07-30
feat(api): add episode consumption creation endpoint

VirUZI:
## Summary

Adds API support for marking a TV episode as watched by creating a new episode consumption through the existing episode history endpoint:

```http
POST /api/v1/media/tv/{source}/{media_id}/{season_number}/{episode_number}/history/
```

The request body is optional. A supplied end_date is used as the watch time; otherwise the current time is used.

Example:
```json
{
  "end_date": "2026-07-29T17:30:00+02:00"
}
```

A successful request returns the newly created consumption using the existing history response format and HTTP 201 Created.

## Motivation

The API previously allowed clients to:

- inspect episode state;
- list episode consumption history;
- update or delete an existing consumption;

but it did not provide a supported way to create an episode consumption.

The generic media creation endpoint is not sufficient for episodes because an
episode belongs to a user-owned season and TV hierarchy and must run Yamtrack's
normal episode, season and TV progress handling.

This endpoint allows API clients to implement a complete episode watch workflow
without reproducing Yamtrack's internal parent-record logic.

## Implementation

Introduces a shared watch_episode operation that:

1. Retrieves and validates provider metadata for the requested season and
episode.
2. Rejects unknown episode numbers without creating invalid records.
3. Reuses an existing season belonging to the authenticated user.
4. Creates missing season and parent TV records when necessary.
5. Preserves the requested media source rather than assuming TMDB.
6. Creates the episode consumption through Season.watch().
7. Runs the existing Episode.save() logic that updates season and TV status.

The existing web episode-save view now uses the same shared operation, avoiding
separate implementations for the UI and API.

Season.watch() now returns the created Episode, allowing the API to
serialize the exact new consumption.

Repeated POST requests are intentionally supported. Each request represents a
new viewing and creates another episode consumption.

API behavior
Success
```http
POST /api/v1/media/tv/tmdb/94997/3/1/history/
Content-Type: application/json

{}
```
Returns:
```http
201 Created
{
  "consumption_id": 123,
  "created": "2026-07-29T15:30:02Z",
  "score": null,
  "progress": 1,
  "progressed_at": "2026-07-29T15:30:02Z",
  "status": 3,
  "start_date": "2026-07-29T15:30:02Z",
  "end_date": "2026-07-29T15:30:00Z",
  "notes": ""
}
```

## Validation
- Invalid media types or sources return 400.
- Invalid end_date values return 400.
- Unknown seasons or episodes return 404.
- Authentication remains required.
- Provider failures use the existing API error-handling conventions.

## Documentation

Updates the OpenAPI schema with:

- the optional end_date request field;
- the 201 response;
- relevant validation, authentication and not-found responses.

## Testing

Adds API coverage for:

- creating an episode consumption;
- preserving a supplied watch time;
- defaulting to the current time;
- reusing an existing season;
- creating missing season and TV parents;
- creating repeat viewings;
- user isolation;
- authentication;
- invalid media types and sources;
- invalid watch times;
- missing episodes;
- provider failures.

The endpoint was also tested from an external API client, including creation of a watch record and subsequent retrieval of the updated episode state.

## AI assistance

AI assistance was used to help draft the initial implementation and tests. I reviewed the changes, tested the endpoint locally, and verified the complete workflow through an external API client.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2847,6 +2847,80 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+    post:
+      tags:
+        - media
+        - history
+      summary: Create an episode watch history record
+      description: Creates another viewing of the requested episode. The request body is optional and end_date defaults to the current timezone-aware time.
+      parameters:
+        - in: path
+          name: media_type
+          required: true
+          schema:
+            $ref: "#/components/schemas/MediaTypes"
+        - in: path
+          name: source
+          required: true
+          schema:
+            $ref: "#/components/schemas/Source"
+        - in: path
+          name: media_id
+          required: true
+          schema:
+            type: string
+        - in: path
+          name: season_number
+          required: true
+          schema:
+            type: integer
+        - in: path
+          name: episode_number
+          required: true
+          schema:
+            type: integer
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                end_date:
+                  type: string
+                  format: date-time
+                  description: When the episode was watched. Defaults to the current time when omitted.
+      responses:
+        "201":
+          description: Created
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ConsumptionInfos"
+        "400":
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "403":
+          description: Forbidden
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "404":
+          description: Episode Not Found
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "500":
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
   /api/v1/media/{media_type}/{source}/{media_id}/{season_number}/{episode_number}/history/{consumption_id}/:
     delete:
       tags:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2852,7 +2852,7 @@ paths:
         - media
         - history
       summary: Create an episode watch history record
-      description: Creates another viewing of the requested episode. The request body is optional and end_date defaults to the current timezone-aware time.
+      description: Creates a new consumption for the requested episode. Every successful request creates another consumption, so repeating the same request records a rewatch rather than updating an existing record. This operation is not idempotent; clients should deduplicate retries if they do not intend to record another viewing. The request body is optional and end_date defaults to the current timezone-aware time.
       parameters:
         - in: path
           name: media_type
--- a/src/api/serializers.py
+++ b/src/api/serializers.py
@@ -504,6 +504,15 @@ class HealthResponseSerializer(serializers.Serializer):
         }
 
 
+class EpisodeWatchSerializer(serializers.Serializer):
+    """Validate an optional watched-at timestamp for a new episode consumption."""
+
+    end_date = serializers.DateTimeField(
+        required=False,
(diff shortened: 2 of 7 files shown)


Cbow2018:
Tested this branch against a fresh instance and it does exactly what the description says. Thanks for building it.

**What I tested**

Built `codex/api-episode-history` as a Docker image, brought it up on a clean database, and POSTed to `/api/v1/media/tv/tmdb/1396/1/1/history/` for a show with nothing tracked at all. That's the case that previously wasn't possible, since `Episode` has no `user` field and hangs off `related_season`.

Result: `201`, and the full hierarchy was created from scratch: episode consumption, Season 1 with the episode under `related.episodes`, and the TV entry. After a second episode, the season reported `tracked: true` with progress 2/7 and status In Progress, and the TV entry rolled up to 2/62. The UI agreed. `Season.watch()` and `Episode.save()` are clearly doing the right thing through this path.

**One behaviour worth confirming**

POSTing the same episode twice creates a second consumption (`consumption_id` 1 and 3 for the same episode). Season and TV progress correctly stayed at 2, counting distinct episodes rather than consumption, so nothing gets corrupted. I assume that's intentional, i.e. a repeat POST is a rewatch rather than an idempotent upsert, but it's worth stating explicitly in the endpoint docs. Automated clients like scrobblers can fire the same request more than once per playback and will need to guard against it themselves.

**Build issue (belongs to `feat/add-api`, not this PR)**

The branch can't be built as a Docker image as-is. `settings.py` lists `rest_framework`, `api` and `drf_spectacular` in `INSTALLED_APPS`, but `pyproject.toml` declares neither `djangorestframework` nor `drf-spectacular`. The build fails at `collectstatic` with `ModuleNotFoundError: No module named 'rest_framework'`. This is true of `FuzzyGrim:feat/add-api` itself, so it's not something introduced here.

Adding the two packages to `pyproject.toml` doesn't fix it on its own, because that invalidates `uv.lock` and `uv sync` then re-resolves everything, at which point it fails trying to fetch `mike` from a git URL in the `docs` group, and the `uv` builder image has no `git` binary.

What worked was leaving the lock untouched and adding one line after it in the Dockerfile:

```
RUN uv sync --locked
RUN uv pip install --python /yamtrack/.venv/bin/python djangorestframework drf-spectacular
```

Proper fix is presumably adding both to `pyproject.toml` and regenerating `uv.lock`. Happy to open a separate PR against `feat/add-api` for that if it'd help.

**On #1606**

On this branch the season detail response already returns tracked episodes with their `consumption_id` under `related.episodes`, so that fix may be partly covered here already. Worth checking before both land.

Context for why I care: I'm looking at adding Yamtrack as a tracker in a Stremio addon, and this endpoint is the missing piece. Without it there's no way for an external client to record a watch.

Cbow2018:
Followed up on the build issue I mentioned: opened #1658  against `feat/add-api` adding `djangorestframework` and `drf-spectacular` to `pyproject.toml` with a regenerated `uv.lock`. Verified by building the image from an unmodified Dockerfile.

Nothing for you to change here, it's a gap in the base branch rather than anything this PR introduced. It just means anyone wanting to test this one can now build it without patching the Dockerfile first.

VirUZI:
@Cbow2018  Thanks for the thorough validation, especially the clean-database and progress roll-up checks.

Repeat POSTs are intentional: each successful request creates a new consumption, so posting the same episode again represents a rewatch rather than an idempotent upsert. I’ll make that—and the need for clients to deduplicate retries—more explicit in the endpoint docs.

The duplicate-delivery concern isn’t specific to this endpoint: other consumption-creation POSTs are also non-idempotent, while action endpoints currently have endpoint-specific behavior. I’ll make the semantics explicit here. A reusable Idempotency-Key mechanism would be the cleaner solution for scrobblers and other retrying clients, but that should be handled separately across the API.

Thanks also for following up with #1658. That build issue is a base-branch dependency/lockfile gap, not introduced here, and is already covered there, so this PR doesn’t need to address it.

Issue #1550, opened 2026-07-04
[BUG] Bug Report: Emby webhook fails to sync watched episodes for Makinavaja: La serie


SOGitHubOS:
# Bug Report: Emby webhook fails to sync watched episodes for Makinavaja: La serie

## Environment

* Yamtrack version: v0.25.3
* Installation: Docker (`ghcr.io/fuzzygrim/yamtrack`)
* Media server: Emby
* Tracking via Emby webhook

## Description

Watched status is correctly detected and logged by Emby, and Yamtrack receives the webhook successfully.

The problem only affects a specific TV series:

* Series: Makinavaja: La serie
* TMDB Series ID: 78198
* TVDB Series ID: 84724
* IMDb Series ID: tt0338623

Other TV series sync correctly through the same Emby/Yamtrack installation.

## Expected behavior

When an episode is marked as watched in Emby, Yamtrack should identify the series and update the tracked status.

## Actual behavior

Yamtrack receives the webhook but fails to resolve the episode to a TMDB entry and aborts processing.

## Relevant logs

Initial test without IMDb episode ID:

```text
[INFO] Extracted IDs from payload: {'tmdb_id': None, 'imdb_id': None, 'tvdb_id': '4103940'}
[INFO] Received webhook for tv: Makinavaja: La serie S01E01
[WARNING] No matching TMDB ID found for TVDB episode ID: 4103940
[WARNING] No TVDB or IMDB ID found for TV episode
```

After manually adding an IMDb ID to the episode metadata:

```text
[INFO] Extracted IDs from payload: {'tmdb_id': None, 'imdb_id': 'tt0896633', 'tvdb_id': '4103940'}
[INFO] Received webhook for tv: Makinavaja: La serie S01E01
[WARNING] No matching TMDB ID found for TVDB episode ID: 4103940
[WARNING] No matching TMDB ID found for IMDB ID: tt0896633
```

## Comparison with a working series

Example from another series that syncs correctly:

```text
[INFO] Extracted IDs from payload: {'tmdb_id': None, 'imdb_id': 'tt0541295', 'tvdb_id': '241540'}
[INFO] Received webhook for tv: Cimarron S01E01
[INFO] Detected TV episode via TMDB ID: 11988, Season: 1, Episode: 1
```

## Additional observations

Yamtrack already appears to know the TMDB series ID because the UI successfully serves:

```text
/details/tmdb/tv/78198/makinavaja-la-serie/season/1
```

[...]

PR #1627, opened 2026-08-06
fix: resolve Plex TV episodes from a TMDB-only episode GUID

v8eta:
Supersedes #1109 (rebased onto current `dev`, with tests). Refs #1071, and the same class of problem discussed in #1550.

## Problem

Plex's `tv.plex.agents.series` agent frequently matches an episode against TMDB **without** a TVDB or IMDB cross-reference, leaving a lone episode-level `tmdb://` GUID. `_process_tv` resolves `anidb -> tvdb -> imdb` only, so the `tmdb_id` that `_extract_external_ids` already returns is discarded and the scrobble is dropped:

```
[INFO]    Extracted IDs from payload: {'tmdb_id': '1347510', 'imdb_id': None, 'tvdb_id': None, 'anidb_id': None}
[INFO]    Received webhook for tv: Below Deck Mediterranean S02E14
[WARNING] No TVDB or IMDB ID found for TV episode
```

This is **per-episode, not per-library**, which is why it reads as "randomly fails" (#1071). Same show, same library, same agent:

| Episode | Plex `Guid` | Result |
|---|---|---|
| S02E06 | `imdb://tt6978246`, `tmdb://1322600`, `tvdb://6113033` | tracked |
| S02E08 | `imdb://tt7093454`, `tmdb://1330649`, `tvdb://6127613` | tracked |
| S02E12 | `tmdb://1336946` | **dropped** |
| S02E14 | `tmdb://1347510` | **dropped** |

The show-level GUIDs are complete; TMDB simply has no episode-level external IDs for those episodes, so re-matching in Plex does not help.

Over one deployment's history this lost **66 of 84** TV webhooks (79%). Movies were unaffected (24/24) because the Plex movie agent supplies all three GUIDs. The webhook returns HTTP 200 throughout, so nothing external shows a fault.

## Approach

TMDB exposes no endpoint that maps an episode ID back to its show. `_process_tmdb_episode` therefore searches by the show title from the payload, then accepts a candidate **only when the episode ID round-trips** against that candidate's season metadata:

```python
for episode in season_metadata.get("episodes", []):
    if str(episode.get("id")) != str(tmdb_episode_id):
        continue
    ...
```

The ID match is what makes the title search safe — a bare title search would be a guess. The episode number is then taken from TMDB rather than the payload index. It runs last, after `tvdb` and `imdb`, so no existing resolution path changes behaviour.

This is the same idea as @Floydian-dk's #1109; that PR was written against the older layout where `_process_tv` still lived in `base.py`, so it no longer applies. Differences here: reads `season_metadata["episodes"]` directly instead of changing the shape of `tmdb.episode()`'s return dict, bounds the search to the first 5 candidates, catches `ProviderAPIError`/`RequestException` rather than bare `Exception`, and adds tests.

## Changes

- `webhooks/tv.py` — `_process_tmdb_episode()`. Also extracts the AniDB branch into `_process_anidb_episode()` so `_process_tv` stays within the project's own `C901`/`PLR0912` limits.
- `webhooks/base.py` — `_get_show_title()` / `_get_season_number()` hooks defaulting to `None`, so Jellyfin and Emby are unchanged until they implement them.
- `webhooks/plex.py` — those two hooks, reading `grandparentTitle` / `parentIndex`.
- `tests/test_webhooks_plex.py` — a TMDB-only GUID that should track, and one whose episode ID matches no candidate and should track nothing.

## Verification

`pytest test_webhooks_plex.py test_webhooks_jellyfin.py test_webhooks_emby.py` — **52 passed**. `ruff check src/integrations/` clean.

Running live, the payload that previously dropped now resolves:

```
[INFO] Detected TV episode via TMDB episode ID: 1347510, TMDB ID: 66902, Season: 2, Episode: 14
[INFO] Marked episode as played: Below Deck Mediterranean S02E14
```

## Not addressed

Episodes whose Plex season number does not exist on TMDB (Plex numbers Shark Week by year, `S2026`) still fail with `Season N not found in The Movie Database`. That is a separate problem.

### DIFF
--- a/src/integrations/webhooks/tv.py
+++ b/src/integrations/webhooks/tv.py
@@ -1,14 +1,19 @@
 import logging
 
+import requests
 from django.utils import timezone
 
 import app
 from app.models import MediaTypes, Sources, Status
 from app.providers import tvdb as tvdb_provider
+from app.providers.services import ProviderAPIError
 
 from . import anime_mappings
 
 logger = logging.getLogger(__name__)
+
+# How many TMDB title-search candidates to confirm an episode ID against.
+TMDB_SEARCH_CANDIDATES = 5
 
 
 class TVWebhookMixin:
@@ -16,44 +21,12 @@ class TVWebhookMixin:
 
     def _process_tv(self, payload, user, ids):
         anidb_id = ids.get("anidb_id")
-        if user.anime_enabled and anidb_id:
-            mapping_data = anime_mappings.fetch_mapping_data()
-            episode_number = self._get_episode_number(payload)
-            mal_id = None
-            mal_episode_number = None
-
-            if not episode_number:
-                logger.warning(
-                    "No episode number found for AniDB ID: %s",
-                    anidb_id,
-                )
-            else:
-                mal_id, mal_episode_number = anime_mappings.get_mal_id_from_anidb(
-                    mapping_data,
-                    anidb_id,
-                    episode_number,
-                )
-
-            if episode_number and not mal_id:
-                logger.info(
-                    "AniDB ID %s not found in mapping, "
-                    "falling through to TV processing",
-                    anidb_id,
-                )
-            elif episode_number:
-                logger.info(
-                    "Detected anime via AniDB ID: %s. Matching MAL ID: %s, Episode: %d",
-                    anidb_id,
-                    mal_id,
-                    mal_episode_number,
-                )
-                self._handle_anime(
-                    mal_id,
-                    mal_episode_number,
-                    payload,
-                    user,
-                )
-                return
+        if (
+            user.anime_enabled
+            and anidb_id
+            and self._process_anidb_episode(anidb_id, payload, user)
+        ):
+            return
 
         tvdb_episode_id = ids.get("tvdb_id")
         if tvdb_episode_id:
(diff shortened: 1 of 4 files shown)


Cbow2018:
Cross-linking #1659, which lands in the same place from the other direction: yours resolves an episode-level `tmdb://`, mine takes a show-level TMDB ID plus the payload's season and episode numbers.

Both add `_process_tmdb_episode` and `_process_anidb_episode` to `tv.py` and both add `_get_season_number`, so they'll conflict whichever merges first. Mine also runs before yours would and treats `tmdb_id` as a show ID unconditionally, which would misfire on the episode-only GUIDs you're targeting.

Happy to rebase onto yours if this one goes first.

src/integrations/webhooks/tv.py
```diff
@@ -74,11 +47,130 @@ def _process_tv(self, payload, user, ids):
...
+            return False
+
+        for candidate in search_results[:TMDB_SEARCH_CANDIDATES]:
+            media_id = candidate["media_id"]
+            try:
+                tv_metadata = app.providers.tmdb.tv_with_seasons(
+                    media_id,
+                    [season_number],
+                )
+            except (ProviderAPIError, requests.exceptions.RequestException, KeyError):
```
FuzzyGrim, line 138: I would remove `KeyError` from here to not swallow real bugs. A non existent season raises `ProviderAPIError`.

src/integrations/webhooks/tv.py
```diff
@@ -74,11 +47,130 @@ def _process_tv(self, payload, user, ids):
...
+                episode_number = episode["episode_number"]
+                logger.info(
+                    "Detected TV episode via TMDB episode ID: %s, "
+                    "TMDB ID: %s, Season: %d, Episode: %d",
+                    tmdb_episode_id,
+                    media_id,
+                    season_number,
+                    episode_number,
+                )
+                self._handle_tv_episode(
```
FuzzyGrim, line 163: This route goes straight to _handle_tv_episode without consulting the anime mappings, so for a user with anime enabled an anime episode rescued here gets tracked as regular TV rather than mapped to MAL.

PR #1652, opened 2026-08-11
Fix anime not getting new episode release date

taigi100:
Found a bug where some animes were not getting newly announced future episodes. I attempted a fix by checking the AniList every time. Not sure if this is a good fix / correct fix, however it seems to work locally


FuzzyGrim:
In a previous version of Yamtrack, there was an issue that prevented new episodes from being downloaded for anime that were already being tracked. This issue should be fixed for newly added anime, and for existing ones, you need to click on Sync Metadata:
image

taigi100:
Ohhh... didn't realize. Thank you - it seems to have worked :)

PR #1658, opened 2026-08-13
fix: add missing API dependencies to pyproject

Cbow2018:
`src/config/settings.py` lists `rest_framework`, `api` and `drf_spectacular` in `INSTALLED_APPS`, but `pyproject.toml` declares neither `djangorestframework` nor `drf-spectacular`. Building the Docker image from this branch fails at `collectstatic`:

This adds both to `dependencies` and regenerates `uv.lock`.

**Verification**

Built the image from an unmodified Dockerfile on this branch. `uv sync --locked` and `collectstatic` both succeed, and the resulting container serves `/api/v1/` and `/api/docs/` normally.

**Note on the workaround**

Adding the packages without regenerating the lock doesn't work: `uv sync` then re-resolves the whole tree and fails fetching `mike` from a git URL in the `docs` group, since the `uv` builder image has no `git` binary. Regenerating `uv.lock` avoids that entirely, which is what this PR does.

Versions match what currently resolves: `djangorestframework~=3.18.0` and `drf-spectacular~=0.30.0`.


66Bunz:
The support for drf spectacular is being added in #1378

Cbow2018:
Thanks, hadn't spotted #1378; closing this as redundant since that branch already declares both

PR #1659, opened 2026-08-13
feat(webhooks): use the payload TMDB ID for TV episodes

Cbow2018:
### Problem

`_process_tv` resolves an episode through AniDB, TVDB, then IMDB. All three need an *episode-level* identifier. A payload that carries only the show's TMDB ID plus `ParentIndexNumber` and `IndexNumber` is dropped with `No TVDB or IMDB ID found for TV episode`, even though `_extract_external_ids` already pulled `tmdb_id` out and the season and episode numbers are sitting in the payload.

The movie handler doesn't have this gap — it treats the payload TMDB ID as its primary route and logs `Detected movie via TMDB ID`. This brings TV into line.

It also affects the IMDB route in practice: a source sending the *series* IMDB ID rather than the episode's lands in `tv_results` instead of `tv_episode_results`, so `_find_tv_media_id` returns nothing and the event is dropped despite TMDB, season and episode all being present.

### Changes

Two commits, separated so the refactor can be reviewed independently.

**1. `refactor(webhooks): extract the AniDB route from _process_tv`**

The AniDB path was inlined while TVDB and IMDB were already separate methods. Extracting it to `_process_anidb_episode` makes the routes consistent and keeps `_process_tv` within ruff's complexity limits once a fourth route is added. No behaviour change — the only difference is that `fetch_mapping_data()` now runs after the episode-number check rather than before it, avoiding a lookup whose result was discarded.

**2. `feat(webhooks): use the payload TMDB ID for TV episodes`**

Adds `_process_tmdb_episode`, placed *after* the existing routes so it is only reached once they have failed. Any payload that works today takes exactly the same path it did before. Also adds `_get_season_number` to the three processors, which each already read the season number inline for logging but never exposed it.

### Testing

Two tests added to `test_webhooks_jellyfin.py`: one for a TMDB-only payload, one where the IMDB lookup misses and TMDB picks it up. All 52 tests across the Jellyfin, Plex and Emby suites pass, and `ruff check` is clean.

Also verified end to end against a running instance. A payload with `{"Tmdb": "1405"}` and no IMDB ID created the show, seasons and episode correctly, and a payload carrying a series-level IMDB ID alongside TMDB fell through to the new route and tracked as expected.

### Motivation

I'm looking at Stremio as a webhook source. Stremio identifies episodes as `tt0903747:1:3` — series IMDB ID plus season and episode — so it can satisfy none of the three existing routes despite holding everything needed to identify the episode.

### DIFF
--- a/src/integrations/tests/test_webhooks_jellyfin.py
+++ b/src/integrations/tests/test_webhooks_jellyfin.py
@@ -263,6 +263,82 @@ class JellyfinWebhookTests(TestCase):
             self.user,
         )
 
+    @patch("integrations.webhooks.base.BaseWebhookProcessor._handle_tv_episode")
+    def test_tv_episode_uses_tmdb_when_no_episode_level_id(
+        self,
+        mock_handle_tv_episode,
+    ):
+        """Test the show's TMDB ID is used when no episode-level ID is sent."""
+        self.user.anime_enabled = False
+        self.user.save(update_fields=["anime_enabled"])
+        payload = {
+            "Event": "Stop",
+            "Item": {
+                "Type": "Episode",
+                "Name": "Episode",
+                "ProviderIds": {"Tmdb": "1396"},
+                "SeriesName": "Breaking Bad",
+                "ParentIndexNumber": 1,
+                "IndexNumber": 3,
+                "UserData": {"Played": True},
+            },
+        }
+
+        response = self.client.post(
+            self.url,
+            data=json.dumps(payload),
+            content_type="application/json",
+        )
+
+        self.assertEqual(response.status_code, 200)
+        mock_handle_tv_episode.assert_called_once_with(
+            "1396",
+            1,
+            3,
+            payload,
+            self.user,
+        )
+
+    @patch("integrations.webhooks.base.BaseWebhookProcessor._handle_tv_episode")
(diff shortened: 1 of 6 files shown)


FuzzyGrim:
Thanks for the PR, but was this actually tested E2E? All your code is based on the assumption that the ids in `Item.ProviderIds` is a series id, but they are episodes id.

A possible solution, which only would work with the unofficial jellyfin webhook plugin, is to get the TMDB show ID in it's `Series.ProviderIds` section.

Cbow2018:
Thanks for the review @FuzzyGrim — the catch was correct.

For the record: on a Jellyfin `Episode` payload, `Item.ProviderIds` holds the **episode** TMDB ID, not the series ID. The series ID is in `Series.ProviderIds`. My implementation read the wrong field.

Closing this in favour of #1627. Both touch the same resolution path so they'd conflict either way, and #1627 is the better base.

Worth noting the two aren't equivalent in scope. #1627 resolves an episode-level `tmdb://` identifier; it doesn't handle a show-level TMDB ID plus the payload's season and episode numbers. Happy to open a separate issue tracking that case so it doesn't get lost, if you think it's worth having on record — otherwise this comment can serve as the note.

Thanks for the time on the review.

PR #1664, opened 2026-08-15
feat: add rewatch support

beschoenen:
This PR allows TV shows to be "rewatched". This means resettings the progress of a TV show without having to delete your play history.

It works by keeping track of a "rewatch_since" nullable DateTime field. When the field is has a value all plays before the date are filtered from the watch progress. Users can remove the value stop a rewatch or change the value to start another rewatch.

Specials seasons are intentionally not included in a rewatch.


omfgnuts:
Can you get into details how this works? @beschoenen

This problem is quite complex and deserves more than three sentences of explanation

If I started series rewatch and watched only 1 episode from season 1 and 1 episode of season 2, what is the status and rewatch status? Does it do seasons rewatch, or episodes, or per tv series?

There are infinite amount of edge cases with rewatching tv series. Please, detail your solution in more detail

PR #1669, opened 2026-08-16
fix: streak shows 0 for unfinished day without plays

beschoenen:
Currently Yamtrack shows 0 in the statistics when the current day has no history entries. The streak should only show 0 when the previous day has no history.

### DIFF
--- a/src/app/statistics.py
+++ b/src/app/statistics.py
@@ -669,8 +669,10 @@ def calculate_streaks(date_counts, end_date):
     longest_streak = 1
     streak_count = 1
 
-    # Check if the most recent active date is today/end_date
-    is_current = active_dates[0] == end_date
+    # Today is still in progress, so it doesn't break a streak yet. end_date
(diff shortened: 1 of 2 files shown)


src/app/statistics.py
```diff
@@ -669,8 +669,9 @@ def calculate_streaks(date_counts, end_date):
     longest_streak = 1
     streak_count = 1
 
-    # Check if the most recent active date is today/end_date
-    is_current = active_dates[0] == end_date
+    # The day in progress doesn't break a streak yet, so activity on either
+    # today/end_date or the day before keeps the current streak alive
+    is_current = 0 <= (end_date - active_dates[0]).days <= 1
```
FuzzyGrim: `end_date` here is not necessarily today, there's a date-range picker

FuzzyGrim (commented):
Thanks for the PR, one thing to sort out before merge: `calculate_streaks` doesn't only ever receive today as `end_date`.

Issue #1279, opened 2026-03-26
[Feature Request] Import from Letterboxd


LazyHatGuy:
**Describe the solution you'd like**  
Being able to import either manually or automatically from Letterboxd would be nice.
I know they offer exporting to a zip file for manual importing

ghost:
I thought it could be simple - it's not.
- https://letterboxd.com/api-beta/ is not available to use, so it's likely a no go.
- RSS feeds only return recent data

So the only viable route is imports from Letterboxd manual user exports.

The zips contain:
- `diary.csv` — every logged watch, with rating, rewatch flag, and tags
- `ratings.csv` — every film you've ever rated (regardless of diary)
- `watched.csv` — every film you've ever marked watched (regardless of diary)
- `watchlist.csv` — films on the watchlist

These overlap heavily: diary entries also appear in `ratings.csv` (if rated)
and `watched.csv`. Ratings entries also appear in `watched.csv`. `watched.csv`
is effectively a superset of everything you've logged or rated.
`watchlist.csv` and `watched.csv` have identical column headers.

`ratings.csv` has films with a score but no watched date. What status should
these get? There's no "rated only" concept in Yamtrack.

The question is if we want to import all of them, or only focus on diary. 

Some behavioural things:
- Create history item on each rewatch in the diary - I think yes - ` bulk_create_with_history` does not support it.
   - but keep latest/oldest dates?
- Upload watchlist - just mark them as planned.
- Backfill `ratings.csv` and `watched.csv`? I don't know - adds a lot of bloat, but some people may be doing those without adding them to their diaries.
- Upload lists - simple to do, but adds bloat. Won't be able to match easily by name after import.
- What happens if TMDB does not match against anything?
- TV series? Probably don't add? Letterboxd exports don't dinstinguish between the two at the moment, but could be added later I guess.

Binocularbath:
I'd be real keen on this. But based on what @wirerat is saying it may not be feasible

ghost:
> I'd be real keen on this. But based on what [@wirerat](https://github.com/wirerat) is saying it may not be feasible

It's possible. It's not going to be perfect - we will have to make trade-offs and can't build everything at once so it's in manageable chunks to review and test. I just put the message out there to find out what we should build first.

Binocularbath:
> > I'd be real keen on this. But based on what [@wirerat](https://github.com/wirerat) is saying it may not be feasible
> 
[...]

PR #1673, opened 2026-08-19
Letterboxd Import

largereptile:
References #1279

This PR adds two new options to the "imports" section of the profile. It allows for any combination of 4 csvs from the letterboxd data export .zip, as well as the scheduled import of films from the letterboxd profile RSS feed.

image

The behaviour should be consistent with other import modules, except unrated films aren't marked as planning. The main issue I've run into is that when importing from a csv there's no TMDB ID and so sometimes it can find the wrong film, but this is an issue with other imports too. It uses primary_release_year to try and reduce this but if a limited TV series on Letterboxd shares the name of a film that released that year (unlikely but possible) it will require manual fixing, as the Letterboxd CSVs don't contain any info about the film apart from the title and year.

Changes outside of adding Letterboxd to standard views/apps/integrations/template files:
- pyproject.toml
  - added import for pandas, used for combining multiple csvs into one standard format.

- src/app/providers/tmdb.py
  - added an optional parameter to allow for requesting a specific year in a request to the tmdb api

- src/integrations/imports/helpers.py
  - added a function to allow getting duplicate media in an existing media check. this allows rewatches to be added from the letterboxd diary/rss uniquely. without duplicate media the check for existing was only on the most recent, and so old watches would be added multiple times.


FuzzyGrim:
What's the difference the csv import and username import? Do we need both?

largereptile:
the main difference is that username import can be done periodically but only has the last~50 or so films, whereas csv import has the longer-term history. if you've only just made your letterboxd account then you can get away with just username, but if you want to shift letterboxd over then keep using it and syncing like mal/etc. then you would use the username import for syncing but the csv for copying everything over

